PR #1, opened 2021-04-06
Lukasarnold/timingtriggercandidate

hep-ml:
translated the timing strain code from triggermodules and triggeralgs to trigger with the necessary changes. "minimal" version.
all logging / error handling is not included yet.
not tested.


hep-ml:
Will create new push request soon

PR #2, opened 2021-04-09
Lukasarnold/timingtriggercandidate

hep-ml:
- data products from triggeralgs
- changes on data product element values
- included a timestamp in "algorithm" 
- no logging
- enum type of TriggerCandidateTypes in trigger


plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -1,55 +1,144 @@
 
-#include "trigger/TriggerCandidate.hh"
-#include "trigger/TriggerCandidateMaker.hpp"
+#include "TimingTriggerCandidateMaker.hpp"
 
-using clock = std::chrono::duration<double, std::ratio<1, 50'000'000>>;
+#include <chrono>
+
+using internal_clock = std::chrono::duration<double, std::ratio<1, 50'000'000>>;
```
philiprodrigues, line 6: In ProtoDUNE-II and DUNE FD, the clock will be 62.5 MHz. Since we probably also need to support "old" ProtoDUNE-I data in the meantime, the clock frequency will have to become a configuration parameter

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -1,55 +1,144 @@
...
+
+  triggeralgs::TriggerCandidate candidate;
+  auto now = std::chrono::steady_clock::now();
+
+  candidate.time_start = data.time_stamp - m_map[data.signal_type].first,  // time_start
+  candidate.time_end = data.time_stamp + m_map[data.signal_type].second, // time_end,
+  candidate.time_candidate = data.time_stamp;
+  candidate.detid = detid_list;
+  candidate.type = TriggerCandidateType::kTiming;
+  candidate.algorithm = uint32_t(internal_clock(now.time_since_epoch()).count());
```
philiprodrigues, line 48: I don't understand this line. Why are you setting the algorithm type to the system clock time?

hep-ml:
The mentioned lines of code create a local time stamp for performance metrics. They shouldn't be in the develop branch, will close this and create a new pull request.

hep-ml:
Comments considered -> please check if error messages correct.

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -40,8 +40,8 @@ TimingTriggerCandidateMaker::TimeStampedDataToTriggerCandidate(const triggeralgs
   triggeralgs::TriggerCandidate candidate;
   auto now = std::chrono::steady_clock::now();
 
-  candidate.time_start = data.time_stamp - m_map[data.signal_type].first,  // time_start
-  candidate.time_end = data.time_stamp + m_map[data.signal_type].second, // time_end,
+  candidate.time_start = data.time_stamp - m_detid_offsets_map[data.signal_type].first,  // time_start
```
philiprodrigues, line 43: This will go wrong if `data.signal_type >= m_detid_offsets_map.size()`, so need to check for that case

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -57,38 +57,33 @@ void
...
   auto params = config.get<dunedaq::trigger::timingtriggercandidatemaker::Conf>();
-  //			try {
-  m_map.push_back({ params.s0.time_before, params.s0.time_after });
-  m_map.push_back({ params.s1.time_before, params.s1.time_after });
-  m_map.push_back({ params.s2.time_before, params.s2.time_after });
-  //			} catch(...)  {
-  //				ERS_LOG(get_name() << " unsuccessfully configured");
-  //			}
-  //			ERS_LOG(get_name() << " successfully configured");
+  try {
```
philiprodrigues, line 60: I don't think there's anything in this block that can throw an `ers::Issue`, is there? If there's no chance of throwing, the try/catch block should be removed (or changed to catch an exception that _can_ actually be thrown here)

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -57,38 +57,33 @@ void
...
+  }
 }
 
 void
 TimingTriggerCandidateMaker::do_start(const nlohmann::json&)
 {
   thread_.start_working_thread();
-  // ERS_LOG(get_name() << " successfully started");
+  std::string oss_prog = get_name() + " successfully started";
+  ers::debug(dunedaq::trigger::ProgressUpdate(ERS_HERE, get_name(), oss_prog));
```
philiprodrigues, line 86: I think `ers::debug(issue)` should become `TLOG_DEBUG(level) << issue;` where `level` is 0, 1, 2, .... Higher levels are more verbose.

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -101,41 +96,38 @@ TimingTriggerCandidateMaker::do_stop(const nlohmann::json&)
...
     }
-    ++receivedCount;
 
-    candidates.push_back(TimingTriggerCandidateMaker::TimeStampedDataToTriggerCandidate(data));
+    candidate = TimingTriggerCandidateMaker::TimeStampedDataToTriggerCandidate(data);
 
-    // std::string oss_prog = "Activity received #"+std::to_string(receivedCount);
-    // ers::debug(dunedaq::dunetrigger::ProgressUpdate(ERS_HERE, get_name(), oss_prog));
+    std::string oss_prog = "Activity received.";
+    ers::debug(dunedaq::trigger::ProgressUpdate(ERS_HERE, get_name(), oss_prog));
```
philiprodrigues, line 113: TLOG_DEBUG() again

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -101,41 +96,38 @@ TimingTriggerCandidateMaker::do_stop(const nlohmann::json&)
...
       }
     }
-  }
 
-  // std::ostringstream oss_summ;
-  // oss_summ << ": Exiting do_work() method, received " << receivedCount
-  //	 << " TCs and successfully sent " << sentCount << " TCs. ";
-  // ers::info(dunedaq::dunetrigger::ProgressUpdate(ERS_HERE, get_name(), oss_summ.str()));
+    oss_prog = "Exiting do_work() method, received and successfully sent.";
+    ers::debug(dunedaq::trigger::ProgressUpdate(ERS_HERE, get_name(), oss_prog));
```
philiprodrigues, line 129: TLOG_DEBUG()

plugins/TimingTriggerCandidateMaker.hpp
```diff
@@ -47,7 +50,7 @@ class TimingTriggerCandidateMaker : public dunedaq::appfwk::DAQModule
 
   std::chrono::milliseconds queueTimeout_;
 
-  std::vector<std::pair<int64_t, int64_t>> m_map;
+  std::vector<std::pair<int64_t, int64_t>> m_detid_offsets_map;
```
philiprodrigues, line 53: Now that I think about it, let's make this a `std::map`. That more clearly represents what it's doing, and it means we don't need to have offsets for all values starting at zero (eg, we could store offsets for just 1, 3 and 5)

philiprodrigues (changes requested):
This looks much better, thanks. There's one outstanding comment left from the previous review, and I have a couple of new ones on this set of changes. There's also an "unused variable" compiler warning at TimingTriggerCandidateMaker.cpp:41 which should be fixed too

plugins/TimingTriggerCandidateMaker.cpp
hep-ml: ok

plugins/TimingTriggerCandidateMaker.cpp
hep-ml: ok

plugins/TimingTriggerCandidateMaker.hpp
hep-ml: Ok

hep-ml:
Comments considered & pushed to branch.

philiprodrigues (approved):

PR #4, opened 2021-04-14
Add module level trigger

philiprodrigues:
This PR adds a ModuleLevelTrigger DAQModule, which takes in TriggerCandidates from the timing system and produces a TriggerDecision for each one, subject to available tokens from the DF subsystem. The PR also contains some fake generators to exercise the MLT and a python confgen to run a simple test setup


eflumerf (approved):
I see that we are still hard-coded to APA0, but a brief code review looks good and the code builds in my test area. Functionality-based tests were not run, and this PR does not add any unit tests (!!).

PR #9, opened 2021-04-15
Use `appfwk::queue_inst` function to get queues

philiprodrigues:
The `appfwk::queue_inst` function does the internal name -> queue instance mapping for a module, in a way that's a bit nicer than looping over all of the available queues. This PR uses that function wherever available.


BenLand100:
Great! I thought it felt a bit clunky looping as before. This is much nicer. Merging.

PR #11, opened 2021-04-16
Thiagobezerra/fake tsd gen

thiagojcb:
- Addition of a simple (no configuration nor options) fake timestamp generator.
- Python config generator similar to listrev. It works with daq_applications, but not with nanorc.


CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ daq_codegen(*.jsonnet TEMPLATES Structs.hpp.j2 Nljs.hpp.j2 )
 daq_add_plugin(ModuleLevelTrigger duneDAQModule LINK_LIBRARIES trigger)
 daq_add_plugin(IntervalTriggerCreator duneDAQModule LINK_LIBRARIES trigger TEST)
 daq_add_plugin(IntervalTCCreator duneDAQModule LINK_LIBRARIES trigger TEST)
+daq_add_plugin(FakeTimeStampDataGenerator duneDAQModule LINK_LIBRARIES trigger TEST)
```
philiprodrigues, line 35: Let's make the name of the module match the object it's producing (so, rename `FakeTimeStampDataGenerator` to `FakeTimeStampedDataGenerator`)

test/plugins/CommonIssues.hpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#pragma once
+
+#include "appfwk/DAQModule.hpp"
+#include "ers/Issue.hpp"
+
+#include <string>
+
+#define TRACE_NAME "CandidateMaker" // NOLINT
```
philiprodrigues, line 19: `TRACE_NAME` specifies the name of the current module for the purposes of debugging output, so it shouldn't go here. I think you can just delete it

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+#include "logging/Logging.hpp"
+
+#include <chrono>
+#include <cstdlib>
+#include <thread>
+#include <string>
+#include <vector>
+
+using pd_clock = std::chrono::duration<double, std::ratio<1, 50000000>>;
```
philiprodrigues, line 23: We'll have to handle the 50 MHz clock in PD-I and the 62.5 MHz clock in PD-II/DUNE, so the clock frequency will have to become a configuration parameter. I think that rules out using a `std::chrono::duration` like this (which needs to be defined at compile-time.

test/plugins/FakeTimeStampDataGenerator.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+private:
+  // Commands
+  void do_configure(const nlohmann::json& obj);
+  void do_start(const nlohmann::json& obj);
+  void do_stop(const nlohmann::json& obj);
+  void do_scrap(const nlohmann::json& obj);
+
+  // Threading
+  dunedaq::appfwk::ThreadHelper thread_;
```
philiprodrigues, line 60: The DAQ style guide naming convention for member variables is `m_variable_name`, so this should become `m_thread`. The same applies to the other member variables.

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+  register_command("start", &FakeTimeStampDataGenerator::do_start);
+  register_command("stop",  &FakeTimeStampDataGenerator::do_stop);
+  register_command("scrap", &FakeTimeStampDataGenerator::do_scrap);
+}
+
+void
+FakeTimeStampDataGenerator::init(const nlohmann::json& init_data)
+{
+  TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Entering init() method";
+  auto ini = init_data.get<appfwk::app::ModInit>();
```
philiprodrigues, line 45: I think this block is more clearly written using `appfwk::queue_inst` (from `appfwk/DAQModuleHelper.hpp`). See an example here:

https://github.com/DUNE-DAQ/trigger/blob/e2bdf8e3522a5df12a854bfa940fa08afa98e471/plugins/FakeDataFlow.cpp#L40

You'll need to pick a name for the queue.

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+void
+FakeTimeStampDataGenerator::do_scrap(const nlohmann::json& /*args*/)
+{
+  TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Entering do_scrap() method";
+  TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Exiting do_scrap() method";
+}
+
+ std::vector<triggeralgs::TimeStampedData>
```
philiprodrigues, line 100: Since we always return exactly one `TimeStampedData`, let's make the return type just `TimeStampedData`, not a vector. The function name `GetTimestamp` is a bit misleading, since it gets a TSD, not a timestamp. How about `get_time_stamped_data` (which also follows the `snake_case` convention from the style guide)

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+{
+  std::vector<triggeralgs::TimeStampedData> tsds;
+  triggeralgs::TimeStampedData tsd{};
+
+  int signaltype = rdm_signaltype(generator);
+
+  auto tsd_start_time = std::chrono::steady_clock::now();
+  tsd.time_stamp = (uint64_t)pd_clock(tsd_start_time.time_since_epoch()).count();
+  tsd.signal_type = signaltype;
+  auto now = std::chrono::steady_clock::now();
```
philiprodrigues, line 111: The counter is just meant to increase by one for each TSD produced, not hold a timestamp.

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+  tsd.counter = (uint32_t)pd_clock(now.time_since_epoch()).count();
+
+  //std::cout << "\033[32mtsd.timestamp: " << tsd.time_stamp << "\033[0m  ";
+  std::cout << "\033[32m" << tsd.time_stamp << ", "<< tsd.signal_type << ", "<< tsd.counter << "\033[0m\n";
```
philiprodrigues, line 115: Use `TLOG_DEBUG()` instead of `std::cout`. No colours please!

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+FakeTimeStampDataGenerator::do_work(std::atomic<bool>& running_flag)
+{
+  TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Entering do_work() method";
+  size_t generatedCount = 0;
+  size_t sentCount = 0;
+
+  while (running_flag.load()) 
+  {
+    TLOG_DEBUG(TLVL_GENERATION) << get_name() << ": Start of sleep between sends";
+    std::this_thread::sleep_for(std::chrono::nanoseconds(1000000000));
```
philiprodrigues, line 132: The sleep time should be configurable. I think this is the one place where we _need_ a configuration parameter for the "initial release". (We'll have many more configuration parameters in future of course)

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    std::vector<triggeralgs::TimeStampedData> tsds = GetTimestamp();
+
+    if (tsds.size() == 0) 
+    {
+      std::ostringstream oss_prog;
+      oss_prog << "Last TSDs packet has size 0, continuing!";
+      ers::debug(dunedaq::dunetrigger::ProgressUpdate(ERS_HERE, get_name(), oss_prog.str()));
```
philiprodrigues, line 140: Use `TLOG_DEBUG` instead of `ers::debug`. You can write `TLOG_DEBUG(level) << "a string";` or `TLOG_DEBUG(level) << some_ers_issue;` whichever you prefer

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    generatedCount+=tsds.size();
+
+    std::string thisQueueName = outputQueue_->get_name();
+    bool successfullyWasSent = false;
+    while (!successfullyWasSent && running_flag.load())
+    {
+      TLOG_DEBUG(TLVL_GENERATION) << get_name() << ": Pushing the generated TSD onto queue " << thisQueueName;
+      for (auto const& tsd: tsds)
+      {
```
philiprodrigues, line 157: Check indentation here

test/plugins/FakeTimeStampDataGenerator.cpp
```diff
@@ -0,0 +1,190 @@
...
+	  outputQueue_->push(tsd, queueTimeout_);
+	  successfullyWasSent = true;
+	  ++sentCount;
+	}
+	catch (const dunedaq::appfwk::QueueTimeoutExpired& excpt)
+	{
+	  std::ostringstream oss_warn;
+	  oss_warn << "push to output queue \"" << thisQueueName << "\"";
+	  ers::warning(dunedaq::appfwk::QueueTimeoutExpired(ERS_HERE, get_name(), oss_warn.str(),
+							    std::chrono::duration_cast<std::chrono::milliseconds>(queueTimeout_).count()));
```
philiprodrigues, line 169: `queueTimeout_` is already a `std::chrono::milliseconds`, so the cast is unnecessary

test/plugins/FakeTimeStampDataGenerator.hpp
```diff
@@ -0,0 +1,88 @@
...
+  std::chrono::milliseconds queueTimeout_;
+
+  // Random Generatior
+  std::vector<triggeralgs::TimeStampedData> GetTimestamp();
+  std::default_random_engine generator;
+  std::uniform_int_distribution<int> rdm_signaltype = std::uniform_int_distribution<int>    (0, 2);
+};
+} // namespace trigger
+  /*
+ERS_DECLARE_ISSUE_BASE(trigger,
```
philiprodrigues, line 75: This comment should go away

philiprodrigues (changes requested):
Looks mostly OK. A handful of fairly minor style things to fix

test/plugins/CommonIssues.hpp
hep-ml: There is also an `Issues.hpp` header in `include/trigger`. I merged my issues into this file, maybe do the same for the time stamp emulator?

test/plugins/FakeTimeStampDataGenerator.cpp
thiagojcb: I'm not familiar with chrono library. This way of implementation comes from the first version of the module. I will keep reading about it, but in the meantime, any suggestions on how to make it properly?

philiprodrigues: You'll have to make the clock frequency a configuration parameter via the schema file, and use the frequency when you calculate the timestamp of your TSD. The `TriggerDecisionEmulator` in `trigemu` does something like this. The schema is:

https://github.com/DUNE-DAQ/trigemu/blob/develop/schema/trigemu/triggerdecisionemulator.jsonnet

and the plugin that uses it is 

https://github.com/DUNE-DAQ/trigemu/blob/develop/plugins/TriggerDecisionEmulator.cpp

thiagojcb: I made the frequency a configuration parameter, but still depending on chrono. I would need a bit more time to understand and use TriggerDecisionEmulator as an example.

philiprodrigues (approved):

PR #12, opened 2021-04-16
Update to modified triggeralgs

philiprodrigues:
This PR makes `trigger` build against the modified `triggeralgs` in https://github.com/DUNE-DAQ/triggeralgs/pull/2


dingp (commented):
Verified that this PR builds successfully together with DUNE-DAQ/triggeralgs#2.

PR #13, opened 2021-04-19
Move TriggerCandidateType to triggeralgs

hep-ml:
Move TriggerCandidateType.hpp from trigger to triggeralgs


philiprodrigues (approved):
Looks OK

PR #17, opened 2021-04-22
Improve behaviour at high trigger rates

philiprodrigues:
Make the trigger app work a bit better at high rates. With these changes I can run the "faketsd" app at 100 kHz with no inhibits (using 1000 initial tokens).

Changes:

* Removed a sleep() from the TokenManager loop (and replaced it with a timeout in pop()), which helped reduce the number of inhibits
* TimingTriggerCandidateMaker and ModuleLevelTrigger now drain their input queues after they receive the stop command. This ensures that all the data in-flight end-of-run is processed. PR'ed separately in #16 because it's more general than just high rates.
* Fixed a race condition in ModuleLevelTrigger's use of TokenManager. It was doing:
```cpp
    trigger_decision_q->push(td);
    token_manager->trigger_sent(td.trigger_number);
```
But at high rates (and with the FakeDF), the TokenManager's background thread may receive the "trigger complete" message for td *before* trigger_sent is called. It sees a "trigger complete" for a trigger number it doesn't know and ignores it. Then the td never gets removed from TokenManager's list of outstanding triggers. Putting the two lines in the other order fixes this.

### DIFF
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -177,18 +177,26 @@ ModuleLevelTrigger::send_trigger_decisions()
     }
 
     bool tokens_allow_triggers = m_token_manager->triggers_allowed();
+
     if (!m_paused.load() && tokens_allow_triggers) {
 
       dfmessages::TriggerDecision decision = create_decision(tc);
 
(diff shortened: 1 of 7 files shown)


BenLand100 (approved):
Looks solid, and tests OK locally for me. Merging.

PR #18, opened 2021-04-23
Benland100/randomtrigger

BenLand100:
Very similar to IntervalTCCreator, but supports uniform and poisson random triggers. Also adds fakertrt test app configuration with both a poisson and uniform random trigger feeding TCs to MLT simultaneously. 

Upgraded MLT input queue to multi-producer-multi-consumer in all tests.

Depends on DUNE-DAQ/triggeralgs#5 to build.


plugins/RandomTriggerCandidateMaker.cpp
```diff
@@ -0,0 +1,171 @@
...
+RandomTriggerCandidateMaker::get_interval(std::mt19937 &gen)
+{
+  switch (m_conf.time_distribution) {
+    case randomtriggercandidatemaker::distribution_type::kUniform:
+      return m_conf.trigger_interval_ticks;
+    case randomtriggercandidatemaker::distribution_type::kPoisson:
+      std::exponential_distribution<double> d(1.0/m_conf.trigger_interval_ticks);
+      return (int)(0.5+d(gen));
+  }
+}
```
philiprodrigues, line 132: gcc gives me a warning here:

```
/home/rodrigues/dune/daq/dunedaq-v2.4.0-testrels/trigger/sourcecode/trigger/plugins/RandomTriggerCandidateMaker.cpp: In member function ‘int dunedaq::trigger::RandomTriggerCandidateMaker::get_interval(std::mt19937&)’:
/home/rodrigues/dune/daq/dunedaq-v2.4.0-testrels/trigger/sourcecode/trigger/plugins/RandomTriggerCandidateMaker.cpp:132:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

Could you silence it (with a `default` or something else)?

plugins/RandomTriggerCandidateMaker.cpp
```diff
@@ -0,0 +1,171 @@
...
+RandomTriggerCandidateMaker::send_trigger_candidates()
+{
+  std::mt19937 gen(m_run_number);
+  // Wait for there to be a valid timestamp estimate before we start
+  if (m_timestamp_estimator->wait_for_valid_timestamp(m_running_flag) == TimestampEstimatorBase::kInterrupted) {
+    return;
+  }
+
+  dfmessages::timestamp_t initial_timestamp = m_timestamp_estimator->get_timestamp_estimate();
+  dfmessages::timestamp_t next_trigger_timestamp = initial_timestamp + get_interval(gen);
```
philiprodrigues, line 144: What do you think of rounding the `next_trigger_timestamp` to be divisible by `trigger_interval_ticks` when a uniform interval is used? It's not really a functional change, but it does make the timestamps easier to look at (and to see the interval between maybe?)

philiprodrigues (changes requested):
Looks good. Just two minor things, in separate comments. It looks like `RandomTriggerCandidateMaker` has all the functionality of `IntervalTCCreator`. If that's the case, shall we just delete the latter?

plugins/RandomTriggerCandidateMaker.cpp
BenLand100: Curious, GCC should know all elements of the enum are implemented. I'll have default treated as kUniform with message printed to the log.

plugins/RandomTriggerCandidateMaker.cpp
BenLand100: No strong feelings here. Agreed this is cleaner at a glance for uniform triggers, and adding this shouldn't have a significant impact for other types of triggers.

BenLand100:
`IntervalTCCreator` still has some test functionality that makes more sense to retain in a "test" plugin: "burst" and "repeat" of TCs. Certainly there is duplicated functionality, but I think some logical separation of test and real plugins is still useful.

philiprodrigues:
> `IntervalTCCreator` still has some test functionality that makes more sense to retain in a "test" plugin: "burst" and "repeat" of TCs. Certainly there is duplicated functionality, but I think some logical separation of test and real plugins is still useful.

Ah right I forgot about those. Then I agree that keeping IntervalTCCreator makes sense

plugins/RandomTriggerCandidateMaker.cpp
philiprodrigues: Thanks, that would be a good change

PR #20, opened 2021-05-05
Set component_type in GeoID instances

eflumerf:
This PR is for compatibility with DUNE-DAQ/dataformats#62 (which should be merged first)


philiprodrigues (approved):
LGTM

PR #26, opened 2021-05-24
Load TA and TC maker algorithms from `triggeralgs` via plugins

philiprodrigues:
Adds plugin infrastructure for TA and TC maker algorithms from `triggeralgs`, so they can be loaded by name (passed by configuration). We create the plugin loading instances for each plugin in this package, since `triggeralgs` doesn't depend on `cetlib`, which we use for DAQ plugin loading.

`TriggerActivityMaker` is an example of how to actually load a plugin at runtime, but it's lacking some things we need:
* Ability to handle "TP to TA" and "TA to TC" without code duplication
* Taking TPSet as input, rather than plain TPs. Then passing individual TPs in start-time order to the algorithm
* Handling heartbeats


BenLand100:
Builds fine locally for me, and looks good. Merging.

PR #27, opened 2021-05-25
Add `trigger::Set<T>`

philiprodrigues:
This PR adds the `trigger::Set<T>` class (with aliases `TPSet = Set<TP>` and `TASet = Set<TA>` as a base for creating the data transport functionality for DS

### DIFF
--- a/include/trigger/Set.hpp
+++ b/include/trigger/Set.hpp
@@ -1,10 +1,18 @@
+/**
+ * @file Set.hpp
+ *
+ * This is part of the DUNE DAQ Application Framework, copyright 2020.
+ * Licensing/copyright details are in the COPYING file that you should have
+ * received with this code.
+ */
+
 #ifndef TRIGGER_INCLUDE_TRIGGER_SET_HPP_
 #define TRIGGER_INCLUDE_TRIGGER_SET_HPP_
 
 #include "dataformats/Types.hpp"
 #include "triggeralgs/TriggerActivity.hpp"
 #include "triggeralgs/TriggerPrimitive.hpp"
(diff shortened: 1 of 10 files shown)


include/trigger/Set.hpp
```diff
@@ -0,0 +1,65 @@
...
+using TASet = Set<triggeralgs::TriggerActivity>;
+
+} // namespace dunedaq::trigger
+
+MSGPACK_ADD_ENUM(dunedaq::trigger::TPSet::Type)
+MSGPACK_ADD_ENUM(dunedaq::trigger::TASet::Type)
+
+DUNE_DAQ_SERIALIZE_NON_INTRUSIVE(dunedaq::trigger, TPSet, seqno, type, from_detids, start_time, end_time, objects)
+DUNE_DAQ_SERIALIZE_NON_INTRUSIVE(dunedaq::trigger, TASet, seqno, type, from_detids, start_time, end_time, objects)
+
```
brettviren, line 64: For proper type factoring, these types and their headers should be moved out of `Set.hpp` and into individual headers.  Eg `TPSet.hpp`.

brettviren: Likewise `serialize.hpp` could be factored.  I suggest each `TxSet.hpp` hold the `Set<Tx>` and the `Tx` related part of the current `serialize.hpp`. 

Eg a TC consumer should not "know" about TPs.

philiprodrigues: Thanks for the suggestion - 9d56071 implements this (actually dividing a little more finely than you suggest, so that the TC serialization header can depend on just the TA serialization, not TASet)

PR #30, opened 2021-05-26
Make a generic base for TA TC TD algorithm plugins

BenLand100:
This takes the logic from the `TriggerActivityMaker` and makes a template base class `TriggerGenericMaker` from which these DAQModules are derived:
* `TriggerActivityMaker`
* `TriggerCandidateMaker`
* `TriggerDecisionMaker`


philiprodrigues (commented):
This looks good, thanks. The only possible issue I can see is, does `TriggerGenericMaker` need a virtual dtor, since it's inherited from?

BenLand100:
I suspect `dunedaq::appfwk::DAQModule` has already declared the destructor to be virtual, which applies to all derived classes, but there's no harm in adding an explicit virtual destructor here.

philiprodrigues:
Ah yes, you're right about the dtor virtualness passing through. Thanks for the addition

PR #34, opened 2021-05-27
 Retool TriggerGenericMaker to behave differently for Set<T> input/output

BenLand100:
Introduces a helper class `TriggerGenericWorker` to deal with the partial specialization of the `do_work` method. This allows queues of type `Set<A>` -> `OUT` and `Set<A>` -> `Set<B>` to be handled correctly. Unfortunately C++ does not yet allow partial specialization of methods in a template class, and the helper class is the cleanest workaround for this.

### DIFF
--- a/src/trigger/TriggerGenericMaker.hpp
+++ b/src/trigger/TriggerGenericMaker.hpp
@@ -23,6 +23,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <algorithm>
 
 namespace dunedaq::trigger {
 
@@ -167,7 +168,7 @@ public:
         continue; //nothing to do
       }
   
-      std::vector<OUT> out_vec; // 1 input -> many outputs
+      std::vector<OUT> out_vec; // one input -> many outputs
       try {
         m_parent.m_maker->operator()(in, out_vec);
       } catch (...) { // TODO BJL May 28-2021 can we restrict the possible exceptions triggeralgs might raise?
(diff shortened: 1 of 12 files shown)


philiprodrigues (commented):
There's quite a bit of nearly-the-same code here, which makes me wonder if it's worth rethinking the lack of a `TCSet`. If we had a `TCSet`, then you wouldn't need the `Worker<Set<A>, OUT, MAKER>` specialization (is that correct?). We could then either modify the downstream code to accept `TCSet`s or we could add some small bit of code that takes the `TCSet`s and "explodes" them into their component `TC`s for consumers. Would that be an overall improvement?

src/trigger/TriggerGenericMaker.hpp
```diff
@@ -78,36 +124,178 @@ class TriggerGenericMaker : public dunedaq::appfwk::DAQModule
...
-            outs.pop_back();
-            successfullyWasSent = true;
-            ++sent_count;
-          } catch (const dunedaq::appfwk::QueueTimeoutExpired& excpt) {
-            ers::warning(excpt);
+      try {
+        parent.m_maker->operator()(in, outs);
+      } catch (...) {
+        ers::error(AlgorithmFatalError(ERS_HERE, parent.get_name(), "algorithm_name_placeholder"));
+        continue;
```
philiprodrigues, line 140: If the algorithm throws, I think we should assume it's not going to work any more, and so we should `break` rather than `continue`.  We can then make the ers call an `ers::fatal` (which just increases the severity of the message; it doesn't exit the program like in other logging systems).

BenLand100:
That's right, if everything were a `Set<T>` one of the specializations could be dropped. That said, the 'extra code' in that specialization would just get moved to whatever module does the exploding, and introduce one more queue's worth of overhead, so having this additional specialization might be the cleanest way in the end.

I did shuffle some code around to reduce how much redundant parts there were in the (renamed) `TriggerGenericWorker` so might be worth another look.

philiprodrigues:
Yes, that looks better, and shows how we can factor out common parts between the specializations. Thanks. A couple more individual-line comments coming...

src/trigger/TriggerGenericMaker.hpp
```diff
@@ -72,42 +115,225 @@ class TriggerGenericMaker : public dunedaq::appfwk::DAQModule
...
+  {
+    while (running_flag.load()) {
+      Set<A> in;
+      if (!m_parent.receive(in)) {
+        continue; //nothing to do
+      }
+
+      Set<B> out; // 1 input set -> 1 output set
+      switch (in.type) {
+        case Set<A>::Type::kPayload:
```
philiprodrigues, line 210: I think we need a little more logic here to ensure that `operator()` is called with objects in `time_start` order when the inputs are derived from multiple streams (eg, TAs from multiple APAs)

You can assume that the input queue of sets is time-ordered, one set per stream. That's the way zipper will order the output, ie, like this:

zipper-output

Then you have to "buffer" the input sets until you see one with a larger start time. At that point you can sort the objects from all of the buffered sets and pass them to `operator()`

BenLand100: In that case, I'm not sure I understand the purpose of a `Set<T>`, or the logic for how a `TPSet` should map onto a `TASet`. If we're just passing TP / TA to the algorithms in time order, and the mapping from TPSet to TASet is arbitrary, why group them?

Is the idea that a `Set<T>` represents a non-overlapping global time slice? i.e. a maker should wait for all the TPSet from a non-overlapping slice, call the algorithm for all objects in that slice in time order, then is free to aggregate all the results into a single TASet? 

If that's the case, is there a mechanism to ensure the `start_time` and `end_time` of the `Set<T>` in a group is the same that should be preserved through this TP->TA->TC chain?

philiprodrigues: > In that case, I'm not sure I understand the purpose of a `Set<T>`, or the logic for how a `TPSet` should map onto a `TASet`. If we're just passing TP / TA to the algorithms in time order, and the mapping from TPSet to TASet is arbitrary, why group them?

The original motivation for grouping the TPs is as an optimization to reduce message rate across the network: eg, in ProtoDUNE, there are ~1MHz of TPs per APA, so reducing message rate is worthwhile. For TAs, this argument probably doesn't apply, or at least not as strongly, but "packaging" them in the same way as TPs means we can reuse algorithms like the Zipper.

> Is the idea that a `Set<T>` represents a non-overlapping global time slice? i.e. a maker should wait for all the TPSet from a non-overlapping slice, call the algorithm for all objects in that slice in time order, then is free to aggregate all the results into a single TASet?

Yes, that's right. In the diagram I pasted earlier, the values t0, t1, t2 are the same for each of the input streams (which makes things global) and the intervals [t0,t1], [t1,t2] etc are non-overlapping. 

> If that's the case, is there a mechanism to ensure the `start_time` and `end_time` of the `Set<T>` in a group is the same that should be preserved through this TP->TA->TC chain?

The `TPWindow` algorithm being adapted from ptmp (slide 10 of [my talk](https://indico.fnal.gov/event/47067/contributions/215031/attachments/142900/180529/ds-data-transport-2021-05-11.pdf) ) ensures this - it rearranges TPs between TPSets so that the output TPSet widths/offsets are as given in configuration. (Eg, if given width=100, offset=30, the output TPSets would have [start_time, end_time] equal to [30, 129], [130, 229], [230, 339] etc).

Then the big-picture configuration would be that N streams of TPs each go into a TPWindow instance (all the instances have the same width and offset), with the outputs fed into `TPZipper` to merge them in a time-ordered way, and the output of the `TPZipper` fed to `TriggerGenericMaker`

BenLand100: Thanks, that makes sense. I'll add some logic to the `TriggerGenericWorker` to buffer, sort, and preserve the time bounds.

philiprodrigues (approved):
This looks good to me. Thanks for bearing with me on the iterations

PR #35, opened 2021-06-01
Add live-time counter

philiprodrigues:
This PR adds a `LivetimeCounter` object that keeps track of the amount of time that the MLT was alive to triggers, and the amount of time it was dead (because there were no tokens).

### DIFF
--- a/src/trigger/LivetimeCounter.hpp
+++ b/src/trigger/LivetimeCounter.hpp
@@ -8,27 +8,57 @@
 
 namespace dunedaq::trigger {
 
+/**
+ * @brief LivetimeCounter counts the total time in milliseconds spent in each of the available states
+ *
+ * The current state is set at construction, and can be changed with
+ * set_state(). The accumulated time in a particular state can be
+ * retrieved with get_time(State), and the times spent in all the
+ * states with get_time_map()
+ */
 class LivetimeCounter
 {
 public:
   enum class State {
-    kLive,
-    kDead,
-    kPaused
+    kLive,  // Live to triggers
+    kDead,  // Dead to triggers due to a problem
+    kPaused // Triggers paused (so we are dead to triggers, but intentionally)
   };
 
+  /**
+   * @brief A type to store a time duration in milliseconds
+   */
   using state_time_t = uint64_t;
-  
+
+  /**
+   * @brief Construct a LivetimeCounter in the given state
+   *
+   * Counting the time in the initial state begins immediately
+   */
   LivetimeCounter(State state);
 
   ~LivetimeCounter();
-  
+
+  /**
+   * @brief Set the current state to @a state
+   *
+   * Updates the accumulated time in the previous state and switches the state
+   */
   void set_state(State state);
 
+  /**
+   * @brief Get a map of accumulated time in milliseconds in each state
+   */
   std::map<State, state_time_t> get_time_map();
 
+  /**
+   * @brief Get the accumulated time in milliseconds spent in a particular state
+   */
   state_time_t get_time(State state);
 
+  /**
+   * @brief Get a nicely-formatted string of the time spent in each state
+   */
   std::string get_report_string();
 
   std::string get_state_name(State state) const;


src/trigger/LivetimeCounter.hpp
```diff
@@ -0,0 +1,52 @@
+#ifndef TRIGGER_SRC_TRIGGER_LIVETIMECOUNTER_HPP_
+#define TRIGGER_SRC_TRIGGER_LIVETIMECOUNTER_HPP_
+
```
brettviren, line 3: Some docstring comment would be helpful!

plugins/ModuleLevelTrigger.cpp
```diff
@@ -102,7 +103,9 @@ ModuleLevelTrigger::do_start(const nlohmann::json& startobj)
   m_paused.store(true);
   m_running_flag.store(true);
 
-  m_token_manager.reset(new TokenManager(m_token_source, m_initial_tokens, m_run_number));
+  m_livetime_counter.reset(new LivetimeCounter(LivetimeCounter::State::kPaused));
```
brettviren, line 106: I believe this is inside the `start` command handler so I expect the state to be set to `kLive`.

But, either way, I see one must implicitly assume some overall DAQ state machine in order to know what `k*` to set.  

Modules are supposed to be agnostic to the overall FSM, right?

src/TokenManager.cpp
```diff
@@ -43,6 +46,9 @@ TokenManager::trigger_sent(dfmessages::trigger_number_t trigger_number)
   std::lock_guard<std::mutex> lk(m_open_trigger_decisions_mutex);
   m_open_trigger_decisions.insert(trigger_number);
   m_n_tokens--;
+  if(m_n_tokens.load()==0) {
+    m_livetime_counter->set_state(LivetimeCounter::State::kDead);
+  }
```
brettviren, line 51: I admit to not fully understanding the flow of "tokens" but isn't this "deadness" reachable if the TD rate fluctuates low enough? 

Or, if heartbeats are contributing to "tokens" then (I think) there is a need to couple the upstream HB rate setting and the queue pop timeout here so that the timeout is longer than the mean HB period.

plugins/ModuleLevelTrigger.cpp
philiprodrigues: I'm still not really sure how the overall FSM is supposed to relate (or not) to individual modules. The MLT is definitely the place that knows about whether triggering is live/dead, so I think it's the right place for counting those times.

I copied the behaviour where we're paused after the "start" command from `TriggerDecisionEmulator` (the "resume" command then enables triggers). I don't quite recall the rationale for having it this way, but no one has complained about it, so I'm inclined to keep it this way for the moment.

src/TokenManager.cpp
philiprodrigues: The tokens here are the same as credit-based flow control (IIUC). Tokens are given to the MLT by the DFO, and a token is "spent" by sending a TD, so tokens effectively control the number of triggers in-flight (between "TD sent to DFO" and "DF subsystem finished with trigger record").

At the moment, if the MLT wants to emit a trigger when it has no tokens, it just drops the TD immediately, which means that a token count of zero implies that the whole detector (module?) is dead until a new token arrives.

This only accounts for one source of deadtime (DF can't keep up); there isn't yet any handling for detecting that something upstream is not alive and doing anything about that.

brettviren: Oh, okay.  Yes, I'm very familiar with CBFC :)

  https://brettviren.github.io/zio/flow-broker.html

Credit-based flow control is intended to solve a problem where the throughput of a high rate, reliable stream becomes limited not by the transport bandwidth but by the latency involved in sending a per-message ACK serially (ie, downstream doing recv-msg, process-msg, send-ack, recv-msg...).

There are a few problems with using it here.

The TD make up the lowest rate stream of the entire DAQ so sending a serial ACK would not limit it.  This means the problem CBFC solves is moot.  Not itself a killer but we do need to ask why CBFC rather than something simpler.

CBFC is a one-to-one protocol and so it negates PUB/SUB.  We want a PUB pattern because more than DFO will be interested in TDs.  ETI being one.  I think a "Latest TD Info Screen" consumer on a shift window will also be wanted.

CBFC is also intended to retain order.  I guess it's implemented without order here because the DFO is a point of fan-out of TD info to some number of EBs.  Thus if an EB gets stuffed and never finishes then DFO never returns credit.  But, `N-1` credits are still in play so MLT is none-the-wiser.  If stuffing continues, the credits could slowly disappear until finally MLT notices.  This gives some long period of uncertain liveliness. 


Actually, what I see here now as I type this is that there is a bigger problem which needs a larger-scale solution.

MLT, **as a system**, must do two things: 

1. make TDs

2. book keep "liveliness"

It need not, and I think should not, enact these two very different behaviors in the same component.

Instead we need a an MLT **component** which only produces TD and a **separate** "Liveliness Bookkeeper" (LBK?) component.  The LBK would receive liveliness updates both from the (now far simpler) MLT as well as the DFO (and other major players[1]).  With this pattern, the LBK can explicitly know that its wards are alive or dead with a much better precision.  The upper bound on the precision would be set by the TD-level heartbeat rate.

In fact, the DFO and the (now simpler) MLT can simply forward their TDs (both real payload and HB) to the LBK.  Both MLT and DFO can provide them via a PUB.  The LBK can then perform a delayed correlation between each source to determine times when MLT is dead or when MLT is live but DFO is dead.

[1] We may generalize LBK and its input message type so that it **every layer** of the DAQ may report.  We would want to coalesce and reduce the rate of TP/TA and maybe even TC streams to produce some rates in the same ballpark as the TD stream.  This would allow LBK to perform differential tracing.  Ie, not only could it say "dead" vs "alive" but where the death occurred.

src/LivetimeCounter.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include "logging/Logging.hpp"
+
+#include <sstream>
+
+namespace dunedaq::trigger {
+
+LivetimeCounter::LivetimeCounter(LivetimeCounter::State state)
+  : m_state(state)
+  , m_last_state_change_time(now())
```
brettviren, line 11: Another issue, the system time is not the right measure for live time.  The data clock is.

src/trigger/LivetimeCounter.hpp
philiprodrigues: Done in a0b2e44

philiprodrigues:
Obsoleted by Michal's work

PR #36, opened 2021-06-01
Use updated field types in `triggeralgs`

philiprodrigues:
This PR contains the changes necessary to build against https://github.com/DUNE-DAQ/triggeralgs/pull/13 (which is currently just a draft)


BenLand100:
Looks like this also includes the livetime counter from #35. Going to merge now, so that compatibility with triggeralgs is maintained, pending the changed discussed in the other PR.

philiprodrigues:
Ah you're right. I included `LivetimeCounter` by accident and meant to remove it from this branch, then forgot. I will submit another PR to remove LivetimeCounter pending the discussion in #35

PR #37, opened 2021-06-09
DAQ module that fakes TP-creator heartbeats

alexbooth92:
Addition of a new plugin which generates fake tp-creator heartbeats. OpMon also implemented. The module is currently un-tested.

### DIFF
--- a/plugins/FakeTpCreatorHeartbeatMaker.cpp
+++ b/plugins/FakeTpCreatorHeartbeatMaker.cpp
@@ -1,0 +1,178 @@
+/**
+ * @file FakeTpCreatorHeartbeatMaker.cpp
+ *
+ * This is part of the DUNE DAQ Application Framework, copyright 2021.
+ * Licensing/copyright details are in the COPYING file that you should have
+ * received with this code.
+ */
+
+#include "FakeTpCreatorHeartbeatMaker.hpp"
+
+#include <string>
+
+namespace dunedaq {
+namespace trigger {
+FakeTpCreatorHeartbeatMaker::FakeTpCreatorHeartbeatMaker(const std::string& name)
+  : DAQModule(name)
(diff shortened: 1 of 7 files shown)


philiprodrigues (commented):
This looks good. I don't see any major problems, based on looking at the code. We will have to check against `TriggerGenericMaker` to make sure we agree on the order and format of heartbeat messages, but if any changes are needed, I expect them to be minor. 

One small but annoying thing: we've abbrieviated "trigger primitive" as "TP" elsewhere in the module, so I think we should be consistent here

plugins/FakeTpCreatorHeartbeatMaker.cpp
```diff
@@ -0,0 +1,178 @@
...
+                                                   bool const& is_first_tpset_received)
+{
+  // If it is the first TPSet received, send out a heartbeat.
+  // Else, can assume that the TPSets are already ordered by start_time. Therefore, check only
+  // that the difference between the start_time of the current TPSet and the time that the
+  // last heartbeat was sent out is greater than the specified heartbeat interval.
+  if (is_first_tpset_received)
+    return true;
+  else
+    return last_sent_heartbeat_time + m_heartbeat_interval > current_tpset_start_time;
```
philiprodrigues, line 163: Is the ">" the right way round here? Say `last_sent_heartbeat_time` is 1000 and `m_heartbeat_interval` is 100, then we should send a heartbeat if  `current_tpset_start_time > 1100`, right?

alexbooth92: Yes, you're right. I'll fix that.

alexbooth92:
With regard to your comment about the abbreviation of trigger primitive, are you referring to changing Tp to TP in the module name?

philiprodrigues:
> With regard to your comment about the abbreviation of trigger primitive, are you referring to changing Tp to TP in the module name?

Yes

philiprodrigues (approved):
Looks good to me. Thanks!

PR #41, opened 2021-06-17
Add `Set<T>::origin` and remove `::from_detids`

### LINKED ISSUE #40: Change to `Set<T>` to add `origin` and remove `from_detids`
These are two independent issues but small enough and in the same context to treat together.

The `from_detids` is ill defined and unused and depending on interpretation is redundant with info in `objects` or which may be provided in configuration.

OTOH, the identity of a source aka origin of a `Set<T>` is needed in order to meaningfully aggregate sets from multiple sources.   

In some cases the origin ID may be opaque.  Eg, in zipper, this identity is needed to define logical "streams" from the single input queue.  Other cases, such as TA/TC creators, a semantic meaningful origin may be required.  

After some discussion, the identification of an origin by a `GeoID` seems most reasonable.

This issue will update `Set.hpp` and clear out any uses of `from_detids`.

A larger issue is to set the new `Set<T>::origin` attribute.  Each producer of `Set<T>` is in principle free to set `origin` in any way.  But, most likely the `GeoID` info should be set via configuration and that info had best follow a common path for all uses.  This means a shared `GeoID` schema and shared C++ helper to convert from config object to `GeoID` struct (which is hand written and not codegen'ed).  This commit will try to show the (a) way but some changes are out of scope.

brettviren:
Eventually fixes #40 but is a WIP for now.


brettviren:
Ready for a look.  Note, things related to serialization of `Set<>` with `GeoID` come in with #43.

include/trigger/TASet.hpp
```diff
@@ -21,6 +21,6 @@ using TASet = Set<triggeralgs::TriggerActivity>;
 } // namespace dunedaq::trigger
 
 MSGPACK_ADD_ENUM(dunedaq::trigger::TASet::Type)
-DUNE_DAQ_SERIALIZE_NON_INTRUSIVE(dunedaq::trigger, TASet, seqno, type, from_detids, start_time, end_time, objects)
+DUNE_DAQ_SERIALIZE_NON_INTRUSIVE(dunedaq::trigger, TASet, seqno, type, start_time, end_time, objects)
```
philiprodrigues, line 24: `origin` needs to be added to the list of members to serialize

include/trigger/TPSet.hpp
```diff
@@ -21,6 +21,6 @@ using TPSet = Set<triggeralgs::TriggerPrimitive>;
 } // namespace dunedaq::trigger
 
 MSGPACK_ADD_ENUM(dunedaq::trigger::TPSet::Type)
-DUNE_DAQ_SERIALIZE_NON_INTRUSIVE(dunedaq::trigger, TPSet, seqno, type, from_detids, start_time, end_time, objects)
+DUNE_DAQ_SERIALIZE_NON_INTRUSIVE(dunedaq::trigger, TPSet, seqno, type, start_time, end_time, objects)
```
philiprodrigues, line 24: Add `origin` here too

plugins/TriggerPrimitiveMaker.cpp
```diff
@@ -74,7 +74,7 @@ TriggerPrimitiveMaker::do_configure(const nlohmann::json& obj)
       tpset.end_time = tpset.start_time + m_conf.tpset_time_width;
       tpset.seqno = seqno++;
       tpset.type = TPSet::Type::kPayload;
-      tpset.from_detids = { tp.detid };
+      // fixme: set tpset.origin
```
philiprodrigues, line 77: Let's pick something for `origin` for the moment. How about:

```cpp
system_type = GeoID::SystemType::kDataSelection;
region_id = 0;
element_id = tp.detid;```

If there's a good way to indicate that this is temporary/provisional, we could use that (set `region_id` to 2^16 - 2???)

In the input files we're using (TPC TPs from ProtoDUNE), detid is set (starting at the lowest bit):

* module number  (2 bits) -- 0
* APA number (8 bits) -- 6
* felix link (6 bits) -- 5 (1st file) and 11 (2nd file)

(ie `(module & 0x3) | ((apa &  0xff) << 2) | ((link & 0x3f) << 10)` if I did the math right)

brettviren: `region_id` is set to the "invalid" value by default.  Maybe best to simply keep that until we understand more globally how best to set these bits.

In the end, I think GeoID bits should be set in configuration and not be set in a data-driven way.  

Instead, the module may (should?) apply some (detector specific?) algorithm to compare its configured GeoID to each input datum and warn+drop the input if it is "out of scope".

brettviren:
I set the `SystemType` in the `Set<>` so all will be `kDataSelection`.  `region_id` I leave as invalid for now and `element_id` is take "last one wins" from the individual TP.

There are other TxSet creators to deal with these `GeoID` settings but I think these need not delay this PR.  Once we have some common way to set `GeoID` from config we can update the lot.

@philiprodrigues please check and merge if you are okay with it.

philiprodrigues:
This looks good to me. I think there are a few things to tidy up, but let's go ahead and merge and postpone them until later:

* Use `m_` member variable convention in `TriggerZipper` and `zipper`
* Add jsonnet schema for the `TriggerZipper` config
* Add some tests for the zipper algorithm

PR #43, opened 2021-06-17
Add new zipper

brettviren:
This integrates a new rewrite of PTMP's TPZipper from 

https://github.com/brettviren/zipper

It is contingent on #41 which is also a WIP.  I'll be rebasing this if the other branch needs more work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,24 @@ daq_add_library(TokenManager.cpp LivetimeCounter.cpp LINK_LIBRARIES appfwk::appf
 ##############################################################################
 # Codegen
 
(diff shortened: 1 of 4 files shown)


brettviren:
That check failure seems unrelated.  Anyways, once #41 is in, we can worry.

brettviren:
I had to do a force push because I managed to bring in a whole bunch of other's work which caused infinite conflicts.  I ended up having to resort to manual patch against `develop`.  And, then in updating `develop` locally I get build errors related to msgpack.  So, I think my dev environment has become invalidated.   But the test have passed.  So, please merge unless there is some major blocker and I can do further cleanups after I get back a sane dev env.

philiprodrigues:
This looks good to me: merges correctly locally and the unit test passes

PR #44, opened 2021-06-17
Add prescale maker plugins and faketp_chain app.

BenLand100:
See the `README.md` in `python/trigger/faketp_chain` for more info on faketp_chain app.

Adds a `configure` step to algorithm plugins that requires DUNE-DAQ/triggeralgs#14 to build.


philiprodrigues (approved):
All looks good to me, and runs locally for me

PR #46, opened 2021-06-18
Confgen for running FakeTPCreatorHeartbeatMaker

alexbooth92:
Minor change to FakeTPCreatorHeartbeatMaker following testing. Confgen script written which runs the module between TriggerPrimitiveMaker and TPSetSink. Module behaves as expected. Note that small changes have been made to the checks of the TPSets that come in to TPSetSink:

-- To avoid "TPSets out of order" message for fake heartbeats, changed the condition on from current.start_time <= last.start_time to current.start_time < last.start_time.
-- To avoid "Empty TPSet" message for fake heartbeat, added logic to recognise TPSets of Type kHeartbeat.


philiprodrigues (approved):
Code looks good to me, and the generated configuration runs correctly in nanorc

PR #47, opened 2021-06-23
Added suggestions for how to download sample TPSet files from cernbox

bieryAtFnal:
To help make it a little easier to remember where to find the sample TPSet files, I added some comments to the faketp_chain/README file.  

If these look reasonable, then hopefully we can approve this PR and merge the branch to _develop_, so that everyone can see these suggestions.


BenLand100:
Looks good to me.

PR #48, opened 2021-06-23
Fixes and enhancements for replay functionality

hep-ml:
See Phil's comments:
"1. The loop that reads in the TPs assumes that the input file is sorted by TP start time. We should check that each TP has a start time >= the previous TP; if not, emit an ers::warning and ignore that TP

2. I think the input-reading loop might not add the last TPSet to m_tpsets, because it's looking for a TP from the *next* TPSet window to cause it to "complete" the current TPSet. Mostly this won't matter, but one day it will confuse someone, so we should fix it now.

3. The sleep_until(next_time) call at line 155 might be asked to wait a long time (eg, if there is a big gap between the TPs, or a large slowdown factor is given). If a stop command is issued during this sleep, the module won't respond to it until the sleep is over. "


hep-ml:
@philiprodrigues 
Tests completed.
Please take note that the output has been verified on the TriggerPrimitiveMaker level, and not on the TriggerActivityMaker level.
Test 1 for time sorting: Ok, error changed to warning
Test 2 to include all m_tpsets: Ok without modifications
Test 3 for breaking long wait times: Lambda changed to common loop break.

philiprodrigues (approved):
Code looks OK now, and works in my local tests (checked that out-of-order TPs are ignored with a warning, and that `stop` is handled quickly even when there is a long wait between TPs).

PR #49, opened 2021-06-24
Sample configurations for publishing TPSets between apps.

bieryAtFnal:
I've create confgen files that attempt to demonstrate the sending of TPSets between DAQ applications using the pub/sub network functionality.  These are in a new `python/trigger/tpset_pub_sub` directory.  A TriggerPrimitiveMaker is used in the producer application, and TPSet_Sinks are used in the consumer apps.

I haven't tested these yet since I got bogged down trying to create the TPSetNQ plugin.  And, I'm leaving this PR in the _draft_ state until I learn more.

Two questions:

1. does the TPSetNQ plugin already exist somewhere?  (if so, where?)
2. has someone else already created an example system configuration that demonstrates the sending of TPSets between DAQ processes?  (if so, please provide a pointer to that, and I will scrap this PR)


philiprodrigues:
Thanks for putting this together. We didn't already have an example, so this is useful. The NQ plugins didn't already exist (an oversight on my part), so I've taken the liberty of creating them and pushing to this branch. While doing that, I noticed a mistake in the `nwqueueadapters` README, which I've fixed - that may have been the cause of you getting bogged down.

I also added a change to the configuration to use MsgPack serialization instead of (the default) JSON, since MsgPack is much faster. JSON is a poorly-chosen (by me) default, so I'll fix that in `nwqueueadapters` too.

bieryAtFnal:
Phil, thanks for all of the additions!  I've updated my local sandbox to include your changes on the branch, and I can see TPSets being sent between processes.
I'm going to convert this to a real PR now, since I believe that the changes are in good shape.

PR #50, opened 2021-06-29
seqno fix

philiprodrigues:
This PR fixes `TriggerPrimitiveMaker` to update the TPSet seqno correctly when the input file is looped over multiple times, and updates `TPSetSink` to check the sequence numbers.


BenLand100:
:+1:

PR #51, opened 2021-06-30
Add GeoID origin to Set<T> output of TriggerGenericMaker

BenLand100:
Relies on the derrived class to call set_geoid and read parameters from its configuration file. Only TriggerActivityMaker emits Set<T>, and is augmented to read region and element info from its config.

Also fixes a few misc typos.


philiprodrigues:
Looks good to me. Thanks

PR #52, opened 2021-07-01
Addition of the BufferManager

thiagojcb:
First version of a BufferManager.

- Template buffer with instances for TPSet, TASet and TCSet
- It has a TPSet DAQ Module (tested with minidaqapp).
- Module receives a queue of TPSets, that are added into the buffer. It also receives a data request, and return its associated fragment (this bit still needs to be filled -> TBD how to do it for v2.10).

Tasks to complete:

- [x] Generalise it (TxSet)
- [x] Write warning messages when requested TPSet not found
- [x] Write the part that handles "pending" data requests (i.e. TPSet not arrived in the buffer yet)
- [x] Write a unit test
- [ ] Add the Fragment filling function of the DAQ Module (for v2.10)


test/plugins/BufferCreator.cpp
```diff
@@ -0,0 +1,185 @@
...
+  size_t sentCount = 0;
+
+  auto start_time = std::chrono::steady_clock::now();
+  auto period = std::chrono::nanoseconds(m_sleep_time);
+  auto next_time_step = start_time + period;
+
+  while (running_flag.load()) {
+
+    TLOG_DEBUG(TLVL_GENERATION) << get_name() << ": Start of sleep between input/output";
+    std::this_thread::sleep_until(next_time_step);
```
philiprodrigues, line 123: I don't think you want a sleep here. We want to handle inputs as fast as they arrive, without spinning the CPU at 100% when nothing is arriving. But the pop timeout on the queues does this for us: if no input arrives, pop() will sleep for up to `timeout`, so the CPU isn't spun

test/plugins/BufferCreator.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    trigger::TPSet input_tpset;
+
+    try {
+      m_input_queue_tps->pop(input_tpset, m_queueTimeout);
+      m_buffer->add(input_tpset);
+      ++addedCount;
+    } catch (const dunedaq::appfwk::QueueTimeoutExpired& excpt) {
+      // skip if no tps in the queue
+      continue;
```
philiprodrigues, line 134: We should still check the data request queue even if there aren't more TPSets, so you don't want to `continue` here - just do nothing in the catch block, and go on to checking the data request queue

test/plugins/BufferCreator.cpp
```diff
@@ -0,0 +1,185 @@
...
+  register_command("stop", &BufferCreator::do_stop);
+  register_command("scrap", &BufferCreator::do_scrap);
+}
+
+void
+BufferCreator::init(const nlohmann::json& init_data)
+{
+  TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Entering init() method";
+
+  m_input_queue_tps.reset(new appfwk::DAQSource<trigger::TPSet>(appfwk::queue_inst(init_data, "tpset_source")));
```
philiprodrigues, line 49: May as well use the `*_source_t` and `*_sink_t` aliases you defined in the header file here

src/BufferManager.cpp
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  if(start_time > m_buffer_latest_end_time) //condition (4)
+  {
+    // add warning here saying data requested hasn't arrived in the buffer yet
+    // need to creat a queue of "pending" data request. How?
+    return tpsets_output;
+  }
+
+  for(auto& tps: m_tpset_buffer)
```
philiprodrigues, line 63: For now it's fine to loop over all of the TPSets to find the ones we want. We may in future want to try to speed this up by noting that `std::set` is stored sorted, so we can only loop over the range of TPSets in the window by using `lower_bound` and `upper_bound` to find the relevant bounds. For large buffers, this is likely to make a performance difference

src/BufferManager.cpp
```diff
@@ -0,0 +1,104 @@
...
+  {
+    if( ( (tps.start_time > start_time) && (tps.end_time   < end_time) ) ||   //condition (1)
```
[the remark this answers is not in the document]
thiagojcb: Indeed. I've removed this first condition, since it is also redundant wrt the other two.

test/plugins/BufferCreator.cpp
thiagojcb: True! Sleep was carried out from the other module I use as example to write this one. Removed now.

src/BufferManager.cpp
thiagojcb: Thanks! I've update it accordingly.

src/BufferManager.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+}
+
+bool
+BufferManager::add(trigger::TPSet& tps)
+{
+  if(m_tpset_buffer.size() >= m_buffer_max_size) //delete oldest TPSet if buffer full -> circular buffer
+  {
+    auto firstIt = m_tpset_buffer.begin();
+    m_tpset_buffer.erase(firstIt);
```
philiprodrigues, line 33: I think you need to update `buffer_earliest_start_time` here, right?

thiagojcb: Indeed! Thanks! Fixed.

src/trigger/BufferManager.hpp
```diff
@@ -0,0 +1,164 @@
...
+   */
+  bool add(BSET& txs)
+  {
+    if (m_txset_buffer.size() >=
+        m_buffer_max_size) // delete oldest TxSet if buffer full (and updating earliest start time) -> circular buffer
+    {
+      auto firstIt = m_txset_buffer.begin();
+      m_txset_buffer.erase(firstIt);
+      firstIt++;
+      m_buffer_earliest_start_time = (*firstIt).start_time;
```
alexbooth92, line 58: m_buffer_earliest_start_time = m_txset_buffer.begin().start_time.begin(); 

After the erase() instead of iterator++? One line less, don't know which is quicker.

test/plugins/TPSetBufferCreator.cpp
```diff
@@ -0,0 +1,336 @@
...
+  frag_h.window_begin = input_data_request.window_begin;
+  frag_h.window_end = input_data_request.window_end;
+  frag_h.run_number = input_data_request.run_number;
+  frag_h.element_id = geoid;
+  frag_h.fragment_type = (dataformats::fragment_type_t)dataformats::FragmentType::kTriggerPrimitives;
+  frag_h.sequence_number = input_data_request.sequence_number;
+
+  frag.set_header_fields(frag_h);
+
+  /*  >>> need to fill the fragment with
```
alexbooth92, line 153: Is this bit post 2.8?

test/plugins/TPSetBufferCreator.cpp
```diff
@@ -0,0 +1,336 @@
...
+            continue;
+          }
+          it++;
+        }
+      } // end if(m_dr_on_hold.size())
+
+    } catch (const dunedaq::appfwk::QueueTimeoutExpired& excpt) {
+    }
+
+    // Block that reveives data requests and return fragments from buffer
```
alexbooth92, line 278: Small typo: reveives to reviews?

src/trigger/BufferManager.hpp
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  enum DataRequestOutcome
+  {
+    kEmpty,
+    kLate,
+    kSuccess
+  };
+
+  struct data_request_output
```
alexbooth92, line 76: Styling: should struct name be LikeThis rather than like_this?

alexbooth92 (commented):
Had a look though, everything makes sense and is well commented.

Added a couple of small suggestions and one or two cosmetic comments here and there.

test/plugins/TPSetBufferCreator.cpp
thiagojcb: Yes. It is not settled yet how the fragments should be filled. Returning empty ones for now.

test/plugins/TPSetBufferCreator.cpp
```diff
@@ -0,0 +1,336 @@
...
+    } catch (const dunedaq::appfwk::QueueTimeoutExpired& excpt) {
+      std::ostringstream oss_warn;
+      oss_warn << "push to output queue \"" << thisQueueName << "\"";
+      ers::warning(dunedaq::appfwk::QueueTimeoutExpired(ERS_HERE, get_name(), oss_warn.str(), m_queueTimeout.count()));
+    }
+  } while (!successfullyWasSent && running_flag.load());
+}
+
+void
+TPSetBufferCreator::send_out_fragment(std::unique_ptr<dataformats::Fragment> frag_out)
```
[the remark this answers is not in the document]
thiagojcb: True, but I'm not sure how to set a default value to a std::atomic<bool>& argument in a function. Any suggestion?

test/plugins/TPSetBufferCreator.cpp
```diff
@@ -0,0 +1,336 @@
...
+          // "<<it->second.size();
+          if ((it->first.window_begin < input_tpset.end_time && it->first.window_begin > input_tpset.start_time) ||
+              (it->first.window_end > input_tpset.start_time && it->first.window_end < input_tpset.end_time) ||
+              (it->first.window_end > input_tpset.end_time &&
+               it->first.window_begin < input_tpset.start_time)) { // new tpset is whithin data request windown?
+            it->second.push_back(input_tpset);
+            // TLOG() << "Adding TPSet (sart_time="<<input_tpset.start_time <<" on DR on hold ("<<
+            // it->first.window_begin  <<", "<< it->first.window_end  <<"). TPSet count: "<<it->second.size();
+          }
+          if (it->first.window_end < input_tpset.start_time) { // If more TPSet aren't expected to arrive then push and remove peding data request
```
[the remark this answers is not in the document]
thiagojcb: Fixed, see 5a6d5b9820511087ecdd00a053bc520a9e0e0ea5

src/trigger/BufferManager.hpp
thiagojcb: Not sure which is quicker, but your suggestion make the code cleaner. Thanks! Fixed (ed511cecccd53868dabb87a8fcad02859c0f7a5a).

test/plugins/TPSetBufferCreator.cpp
thiagojcb: Fixed, see 5a6d5b9820511087ecdd00a053bc520a9e0e0ea5

src/trigger/BufferManager.hpp
thiagojcb: Fixed, see 5a6d5b9820511087ecdd00a053bc520a9e0e0ea5

test/plugins/TPSetBufferCreator.cpp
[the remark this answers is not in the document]
philiprodrigues: I think you can do it with an extra layer of indirection. Put the body of the `do...while` loop in its own function, which returns true if the send succeeded, and false otherwise, then call that function in a loop from your two `send_out_fragment` functions

philiprodrigues:
With a small change to make it build, this works in my local tests, so I think this is good to merge. Thanks @thiagojcb and @alexbooth92 !

PR #56, opened 2021-07-06
Use the TC start and end times in the trigger decision

### LINKED ISSUE #55: Would it be possible to make the TriggerDecision ComponentRequest widths configurable?
Hi,
I'm doing some tests of long-window readout code, and it would be convenient if I could change the width of the ComponentRequests that are sent out in the TriggerDecision via configuration.

This would affect the following line of code, I believe:

https://github.com/DUNE-DAQ/trigger/blob/c5e20db37b81889f825ce49940e40fed471bdaaf/plugins/ModuleLevelTrigger.cpp#L161

I'm not asking for anything fancy, just the use of a single config param in the calculation of the request.window_end for all components in the TD.

Thanks!

philiprodrigues:
This fixes #55 by using the trigger candidate start and end times in the trigger decision sent out by `ModuleLevelTrigger`

### DIFF
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -156,9 +156,8 @@ ModuleLevelTrigger::create_decision(const triggeralgs::TriggerCandidate& tc)
   for (auto link : m_links) {
     dfmessages::ComponentRequest request;
     request.component = link;
-    // TODO: set these from some config map
-    request.window_begin = tc.time_candidate;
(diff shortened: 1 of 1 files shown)


bieryAtFnal:
This change seems good to me.
I have run a test in which I used the new long-window-readout code in _dataformats, dfmessages, and dfmodules_ (actually, it's partly the AddSequenceNumber code) along with the code changes from this branch and seen a long window work as expected.

PR #58, opened 2021-07-09
First pass at introducing a triggering algorithm based on TP ADC integral.

alexbooth92:
More details to follow before an actual pull request is made. Would also like to include a README.


alexbooth92:
Tested on both tps_link_*.txt files, the algorithm is behaving as expected. Added a README in triggeralgs/include/triggeralgs/ADCSimpleWindow which discusses the ActivityMaker and CandidateMaker classes that have been developed. There are also details on how to generate a nanorc configuration and run the algorithm.

At a later stage, as advised by Ben L, I would like to add some logic to drain the TAMaker buffer upon receiving a stop command.

PR #62, opened 2021-07-14
Add TimeSliceOutputBuffer to batch Maker output into sets.

BenLand100:
This works by buffering the output of a `TriggerGenericMaker` that emits `Set<B>` until some configurable time after a window before emitting a `Set<B>` for that window. Windows with no output are not emitted. Windows are inferred from arriving `Set<A>`.

To discuss:
* What to do with events flushed from heartbeats (currently just added to the buffer).
* What to do if a `B` is created that does not fit into any window from the input `Set<A>` (is this possible? currently dropped with a warning, like tardy `B`.)

### DIFF
--- a/include/trigger/Issues.hpp
+++ b/include/trigger/Issues.hpp
@@ -59,6 +59,20 @@ ERS_DECLARE_ISSUE_BASE(trigger,
                        "The " << algorithm << " maker encountered Sets with inconsistent start/end times.",
                        ((std::string)name),
                        ((std::string)algorithm))
+                       
+ERS_DECLARE_ISSUE_BASE(trigger,
+                       TardyOutputError,
+                       appfwk::GeneralDAQModuleIssue,
+                       "The " << algorithm << " maker generated a tardy output, which will be dropped.",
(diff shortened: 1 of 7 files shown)


philiprodrigues:
This looks good to me. Let's merge now; if any of the discussion points result in changes needed, we can address them in later PRs.

>     * What to do with events flushed from heartbeats (currently just added to the buffer).

That sounds right to me.
 
>     * What to do if a `B` is created that does not fit into any window from the input `Set<A>` (is this possible? currently dropped with a warning, like tardy `B`.)

I guess it's possible in the sense that an algorithm can set the start time of its outputs to whatever it likes, but any meaningful start time would presumably have to fit in a window from the input. Let's leave it as-is for now, and we can rethink if the problem ever comes up.

Some other things:

* We discussed on Slack the behaviour at stop() - let's make sure that's handled
* Would `TimeSlice(Input|Output)Buffer` be amenable to some unit tests? That would give us some more confidence in making changes in future, especially if someone other than you has to do them.

PR #63, opened 2021-07-20
Increment TP timestamps after the first send, instead of before

philiprodrigues:
When we replay the TPs multiple times, we increment their timestamps
each time, so that the receiver sees always-increasing timestamps
(instead of jumping back to the start each time). The code was
previously doing that increment _before_ sending the TPs, which meant
that they were never sent with their original timestamp values from
the file. Not really a problem, but it made debugging a bit harder, so
now we increment timestamps only after the first send. Then, the first
loop over the TPs sends them out with their origin timestamp values.


alexbooth92:
So the logic which increments the timestamps remains the same, it just now happens after the push rather than before. Makes sense to me and looks OK.

philiprodrigues:
Thanks for checking, Alex

PR #64, opened 2021-07-20
Set the sequence number on the zipper output Set

philiprodrigues:
The sequence numbers on the output Sets from `TriggerZipper` were not being set - this PR adds that. I assume this is what was intended.


brettviren (approved):
This was an unintentional omission.  Thanks for catching and fixing!

PR #65, opened 2021-07-21
Multiple input streams with heartbeats and zipper

philiprodrigues:
This PR modifies the `faketp_to_sink` and `faketp_chain` python confgens to allow multiple input files. Each input stream has a `TriggerPrimitiveMaker` connected to a `FakeTPCreatorHeartbeatMaker` ("ftpchm"); the outputs from all of the "ftpchm"s go into a `TriggerZipper` which in turn feeds into either the `TPSetSink` (for `faketp_to_sink`) or `TriggerActivityMaker` (for `faketp_chain`).

Each additional input file is added by giving another `--input-file/-f` argument to the confgen, eg:

```
python -m trigger.faketp_chain -f tps_link_11_no_noisy.txt -f tps_link_05_no_noisy.txt [...more options]
```

I also removed `faketp_to_sink_faketpcreatorheartbeatmaker`, because its functionality is now all in `faketp_to_sink`.

With `faketp_chain`, I see a lot of messages like:

```
LOG [void <snip> at .../TriggerGenericMaker.hpp:333] Suppressing empty window
```

Are these the heartbeats, @BenLand100 ?


BenLand100:
I'll have a closer look later today or tomorrow, but for now: those messages refer to windows from the input that had no output objects. This is probably very common, and the message could be removed if it's too verbose.

alexbooth92:
- I was able to generate and run a configuration which uses the pre-scale algorithm with multiple input files.
- I was also able to generate a configuration which uses the ADCSimpleWindow algorithm with multiple input files but didn't yet check that I could run with it (that algorithm is in another branch). The generated config looked sensible.
- One comment would be that since I did not specify any configuration for the ADCSimpleWindow CandidateMakerClass, the default configuration parameter is still dumped to the json, namely "prescale: 1000". This doesn't do any actual harm as far as I can see and the previous faketp_chain confgen behaved in the same way, but it might cause confusion.

alexbooth92 (commented):
- I was able to generate and run a configuration which uses the pre-scale algorithm with multiple input files.

- I was also able to generate a configuration which uses the ADCSimpleWindow algorithm with multiple input files but didn't yet check that I could run with it (that algorithm is in another branch). The generated config looked sensible.

- One comment would be that since I did not specify any configuration for the ADCSimpleWindow CandidateMakerClass, the default configuration parameter is still dumped to the json, namely "prescale: 1000". This doesn't do any actual harm as far as I can see and the previous faketp_chain confgen behaved in the same way, but it might cause confusion.

BenLand100 (approved):
Looks good to me. Working fine locally.

BenLand100:
Regarding the prescale config parameters - I put those in as the defaults, since the default prescale algorithms in the test requires them. Since there is no schema (yet) it won't break anything. I'd suggest always including an appropriate dictionary of arguments if you change the algorithms to something non-default.

PR #67, opened 2021-07-22
Improvements to TriggerGenericWorker output set buffering

### LINKED ISSUE #66: buffer_time not set correctly in `TriggerGenericMaker` -> `TriggerGenericWorker` -> `TimeSliceOutputBuffer`
`m_buffer_time` in `TimeSliceOutputBuffer` is set via `TriggerGenericWorker`'s initialization list, here:

https://github.com/DUNE-DAQ/trigger/blob/175e61ab197b8eb3686bfd82cce3c87cdec72d77/src/trigger/TriggerGenericMaker.hpp#L231

TGW's ctor is called from `TriggerGenericMaker`'s ctor, but the buffer time in TGM isn't set until the `do_configure()` function, so `TimeSliceOutputBuffer` ends up reading an uninitialized value.

I fixed it locally by making the `TriggerGenericMaker::worker` member into a `unique_ptr` and only creating it in `do_configure` after calling `make_maker()`, but there are probably other ways to do it

BenLand100:
This PR includes the changes discussed for TGW behavior on output set creation.

- Fixes #66 added `reconfigure` method to TGW to update their state when `do_configure` is called on TGM.
- Adds draining-on-stop to all TGW implementations with a `drain` method. Both buffered input and output are drained. This includes a change of philosophy from retrying outputs that failed to go onto the queue to warning and dropping said outputs (necessary to avoid potential infinite loops, since the stop signal is not a hard stop anymore).
- TimeSliceOutputBuffer now has a `window_time` parameter and assumes output windows start at a multiple of the configured value. Previous behavior mirrored input sets exactly, but not all possible windows were expected on input sets.


philiprodrigues (approved):
I tested this locally with both prescale and "ADC simple window" algorithms, and it looks OK. I see all of the inputs making it through to outputs.

PR #74, opened 2021-08-03
Improve stop/start behaviour in TriggerZipper and TriggerGeneric*

philiprodrigues:
A case we have to be able to handle is taking two runs in one process, ie the daq_application receives commands " ... start 1 stop start 2 stop ...". So that means we have to drain input queues at stop, and reset any relevant state before starting the next run. This PR has some changes to do that in `TriggerZipper` and `TriggerGeneric(Maker|Worker)`. In the latter, I've introduced `reset()` methods which are called at stop, but one could do it other ways.

I've also removed the pause/resume handling from TriggerZipper (the idea of pause/resume is that they're commands issued by a human from, eg the run control, rather than having to do with backpressure, etc, so they just get handled by the MLT).

### DIFF
--- a/plugins/TriggerZipper.hpp
+++ b/plugins/TriggerZipper.hpp
@@ -81,8 +81,6 @@ public:
         register_command("conf",   &TriggerZipper<TSET>::do_configure);
         register_command("start",  &TriggerZipper<TSET>::do_start);
         register_command("stop",   &TriggerZipper<TSET>::do_stop);
-        register_command("pause",  &TriggerZipper<TSET>::do_pause);
-        register_command("resume", &TriggerZipper<TSET>::do_resume);
         register_command("scrap",  &TriggerZipper<TSET>::do_scrap);
     // clang-format on
   }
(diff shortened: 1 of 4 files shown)


philiprodrigues:
I should have said: the upstream DAQ currently resets the timestamp of its fake data at stop (so the timestamps of TPs look like "1 2 3 4 5 stop 1 2 3 4 5"), so we have to deal with that too. It's possible that some of the changes I made are only necessary for dealing with this case.

src/trigger/TriggerGenericMaker.hpp
```diff
@@ -144,17 +144,23 @@ class TriggerGenericMaker : public dunedaq::appfwk::DAQModule
...
+      bool got = receive(in);
+      if (got) {
         worker.process(in);
+      } else {
+        if (!running_flag.load()) {
+          break;
+        }
       }
-    }
+    } // end while(true)
```
BenLand100, line 159: This is the same behavior, right? Or am I missing something subtle?

plugins/TriggerZipper.hpp
```diff
@@ -206,5 +213,5 @@ class TriggerZipper : public dunedaq::appfwk::DAQModule
 
 #endif
 // Local Variables:
-// c-basic-offset: 4
+// c-basic-offset: 2
```
brettviren, line 216: Lawful-evil! :)

brettviren (approved):
Zipper related looks good to me.  Thanks.  

The `m_zm.clear()` in the "stop" handler should be enough to handle the resetting of times expected after a subsequent "start".

PR #76, opened 2021-08-09
Support multiple TP buffers

philiprodrigues:
Add region and element settings so we can have multiple buffers in one app. Has some unrelated changes to logging which you may or may not want


thiagojcb (approved):
Changes look fine. It also compiled without problems locally.

PR #77, opened 2021-08-11
Clear `streams` data structure in `zipper::clear()`

philiprodrigues:
To get stop/start working, we need to also clear the `streams` data structure when we stop. Otherwise we end up with erroneously tardy messages in the subsequent runs. This PR fixes the problem reported by @floriangroetschla and @bieryAtFnal in which trigger rates in the subsequent runs are lower.

I also added an ers::warning when a tardy set is received, and counts of tardy sets per source.


brettviren (approved):
This is very nice.  Thank you, Phil!

PR #79, opened 2021-08-19
Fix std::bad_alloc that can occur in TPSetBufferCreator

eflumerf:
While doing the `dfmodules` six_process_multi_file integration test, I was consistently getting failures due to the trigger process crashing. The cause was a std::bad_alloc originating at line 257 of TPSetBufferCreator.cpp. I have made a few improvements to the code in this area to resolve the issue and improve readability.


philiprodrigues (approved):
I haven't been able to reproduce the bug, which Eric reports only happens on some machines. But the new code looks correct, so approving.

PR #84, opened 2021-12-14
Feature/dfo

eflumerf:
Initial implementation of the DFO, in the trigger process connected via queue to MLT


mroda88 (approved):
All good. I tested the code with a long run with software-TPG enabled and all was stable for more than 18 hours. A snapshot of the trigger shows the inhibiting mechanism works.
Screenshot from 2021-12-15 08-59-10

bieryAtFnal (approved):
Tests were run successfully.

PR #85, opened 2021-12-15
Added try/catch around queue push in TPSetReceiver dispatch method

bieryAtFnal:
While running some stress tests of the system, I noticed a crash of the trigger app that I believe was caused by an uncaught exception when trying to push a received TPSet onto the appropriate queue.  

This change simply adds a try/catch block around the queue push method, and with the change I no longer see the crash.


eflumerf:
What is the appropriate response when this queue is backed up? Should there be some kind of retry logic? I'm inclined to think the answer is "no", since the queue in question has a size of 1000, so if its full and not accepting tpsets, then things are very backed up...

eflumerf (approved):
LGTM

bieryAtFnal:
Good question about the appropriate response to a failure to push, and I'm not sure that I know the answer, but I agree with your sense that retries are probably not the right response.  Of course, this is probably a good topic for a discussion with interested people sometime.

PR #86, opened 2022-01-27
check if run is paused before complaining about inhibitted triggers

bieryAtFnal:
In some of the integration tests that I've run, I've noticed complaints about "trigger is inhibited" even when the run is paused.
That seems backward to me.  And, it generates spurious complaints that confused the integration test logfile checking.
It seems to me that if the run is paused, then we shouldn't complain about inhibits.
This code change reverses the order of the tests when such checks are made.


eflumerf (approved):
LGTM. I'll integrate this change into the `feature/IndependentDFO` branch as well.

PR #88, opened 2022-02-07
Independent DFO

mroda88:
This PR changes trigger to allow the DFO to live in its own application.
The main consequence for the MLT is the change in the communication protocol between DFO and MLT.


philiprodrigues (approved):
The code looks correct to me. I ran at high trigger rate (200Hz with three RUs) with both the `nanorc` and `newconf` configurations in minidaqapp, and I see the trigger being inhibited and the inhibit being lifted in a sensible way

PR #89, opened 2022-02-11
Use trigger objects from detdataformats

philiprodrigues:
This PR updates `trigger` to use the trigger objects defined in detdataformats, via `triggeralgs`. The detdataformats changes are in https://github.com/DUNE-DAQ/detdataformats/pull/9 and the `triggeralgs` changes are in https://github.com/DUNE-DAQ/triggeralgs/pull/22


philiprodrigues:
One slight misfeature is that this PR contains a new unit test for the overlay/trigger object conversion that should really live in triggeralgs, but that package doesn't depend on the boost unit test package (and we're trying to keep its dependencies minimal in the hopes of being able to share it with offline), so I've put the unit test here

PR #98, opened 2022-03-21
Replay multiple TP streams from file

philiprodrigues:
This PR adds functionality to `TriggerPrimitiveMaker` to allow it to replay TP streams from multiple files, while keeping them all (approximately) in sync, based on timestamp.

### DIFF
--- a/plugins/TriggerPrimitiveMaker.cpp
+++ b/plugins/TriggerPrimitiveMaker.cpp
@@ -52,6 +52,12 @@ TriggerPrimitiveMaker::do_configure(const nlohmann::json& obj)
 {
   m_conf = obj.get<triggerprimitivemaker::ConfParams>();
 
+  // For each of the streams that are specified in the config, we read
+  // the input file, and create an outgoing sink. We also keep track
+  // of the total timestamp range of all the streams, so we can keep
+  // the timestamps of the multiple streams in sync when replaying,
+  // even when they don't all start or end at the same time
+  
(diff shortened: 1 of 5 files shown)


plugins/TriggerPrimitiveMaker.cpp
```diff
@@ -28,34 +30,106 @@ namespace dunedaq::trigger {
...
+  TLOG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Exiting do_start() method";
+}
+
+void
+TriggerPrimitiveMaker::do_stop(const nlohmann::json& /*args*/)
+{
+  TLOG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Entering do_stop() method";
+  m_running_flag.store(false);
+  for (auto& thr : m_threads) {
+    thr.join();
```
eflumerf, line 108: When persisting threads in a class, I like to use `std::unique_ptr<std::thread>` so that you can check that the thread was created using `thr != nullptr`. Even if you don't do that, I'd make sure to call `thr.joinable()`

eflumerf:
Code ran fine with provided input tpset files.

plugins/TriggerPrimitiveMaker.cpp
philiprodrigues: Done

eflumerf (approved):
LGTM

philiprodrigues:
Thanks!

PR #100, opened 2022-03-24
HSI trigger type passthrough changes (ICEBERG) 

MRiganSUSX:
Main:
-> Command-line argument `--hsi-trigger-type-passthrough` added which affects the trigger decision type assignment in MLT, and the size of the readout window in the TTCM
Other:
-> in TTCM, candidate.detid is now set to the signal_map of the event, not just the header
-> TTCM now contains a check on the trigger type for HSI events
-> MLT shifts the TC type left by 8 bits for TC type other than kTiming (if the HSI flag is active)
-> test app to read the hdf5 files and print the trigger type (Phil)

### DIFF
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -148,9 +148,17 @@ ModuleLevelTrigger::create_decision(const triggeralgs::TriggerCandidate& tc)
   decision.trigger_number = m_last_trigger_number + 1;
   decision.run_number = m_run_number;
   decision.trigger_timestamp = tc.time_candidate;
-  // TODO: work out what to set this to
-  decision.trigger_type = 1; // m_trigger_type;
   decision.readout_type = dfmessages::ReadoutType::kLocalized;
+
+  // TODO: work out what to set this to
+  decision.trigger_type = (tc.type == triggeralgs::TriggerCandidate::Type::kTiming) ? tc.detid : 1; // m_trigger_type;
+  std::cout << std::endl;
+  std::cout << "!!!!! TESTING MLT !!!!!" << std::endl;
+  std::cout << "candidate detid: " << tc.detid << std::endl;
+  std::cout << "type: " << (int)tc.type << std::endl;
+  std::cout << "trigger type: " << decision.trigger_type << std::endl;
+  std::cout << std::endl << std::endl;
+
 
   for (auto link : m_links) {
     dfmessages::ComponentRequest request;
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -155,10 +155,17 @@ ModuleLevelTrigger::create_decision(const triggeralgs::TriggerCandidate& tc)
 
   // TODO: work out what to set this to
   if (m_hsi_passthrough == true){
-  	TLOG_DEBUG(3) << "HSI PT IS TRUE!";
-  	decision.trigger_type = (tc.type == triggeralgs::TriggerCandidate::Type::kTiming) ? tc.detid : 1;
+    TLOG_DEBUG(3) << "HSI PT IS TRUE!";
+    TLOG_DEBUG(3) << "TCTYPE: " << (int)tc.type;
+    if (tc.type != triggeralgs::TriggerCandidate::Type::kTiming){
+      decision.trigger_type = tc.detid;
+    } else {
+      dfmessages::trigger_type_t trigger_type_shifted = ((int)tc.type << 8);
+      decision.trigger_type = trigger_type_shifted;
+    }
+  TLOG_DEBUG(3) << "DEC TYPE: " << decision.trigger_type;
   } else {
-	decision.trigger_type = 1; // m_trigger_type;
+    decision.trigger_type = 1; // m_trigger_type;
   }
 
   TLOG_DEBUG(3) << "!!!!! TESTING MLT !!!!!";
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -157,7 +157,7 @@ ModuleLevelTrigger::create_decision(const triggeralgs::TriggerCandidate& tc)
   if (m_hsi_passthrough == true){
     TLOG_DEBUG(3) << "HSI PT IS TRUE!";
     TLOG_DEBUG(3) << "TCTYPE: " << (int)tc.type;
-    if (tc.type != triggeralgs::TriggerCandidate::Type::kTiming){
+    if (tc.type == triggeralgs::TriggerCandidate::Type::kTiming){
       decision.trigger_type = tc.detid;
     } else {
       dfmessages::trigger_type_t trigger_type_shifted = ((int)tc.type << 8);
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -85,6 +85,9 @@ ModuleLevelTrigger::do_configure(const nlohmann::json& confobj)
   }
   m_trigger_decision_connection = params.dfo_connection;
   m_inhibit_connection = params.dfo_busy_connection;
+  m_hsi_passthrough = params.hsi_trigger_type_passthrough;
+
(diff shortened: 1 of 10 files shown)


plugins/ModuleLevelTrigger.cpp
```diff
@@ -148,10 +149,22 @@ ModuleLevelTrigger::create_decision(const triggeralgs::TriggerCandidate& tc)
   decision.trigger_number = m_last_trigger_number + 1;
   decision.run_number = m_run_number;
   decision.trigger_timestamp = tc.time_candidate;
-  // TODO: work out what to set this to
-  decision.trigger_type = 1; // m_trigger_type;
   decision.readout_type = dfmessages::ReadoutType::kLocalized;
 
+  // TODO: work out what to set this to
```
philiprodrigues, line 154: I guess this comment is obsolete and can be removed

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -33,17 +33,23 @@ TimingTriggerCandidateMaker::HSIEventToTriggerCandidate(const dfmessages::HSIEve
...
+    candidate.time_end   = data.timestamp + hsi_pt_after;
   } else {
-    throw dunedaq::trigger::SignalTypeError(ERS_HERE, get_name(), data.signal_map);
+    if (m_detid_offsets_map.count(data.signal_map)) {
+      // clang-format off
+      candidate.time_start = data.timestamp - m_detid_offsets_map[data.signal_map].first;  // time_start
+      candidate.time_end   = data.timestamp + m_detid_offsets_map[data.signal_map].second; // time_end,
+      // clang-format on    
+    } else {
+      throw dunedaq::trigger::SignalTypeError(ERS_HERE, get_name(), data.signal_map);
```
philiprodrigues, line 47: I think this can be `ers::error(dunedaq::trigger::SignalTypeError(ERS_HERE, get_name(), data.signal_map));` instead: that will report an error to the logging system without aborting the thread.

test/apps/print_trigger_type.cxx
```diff
@@ -0,0 +1,39 @@
...
+  std::string filename;
+  app.add_option("-f,--file", filename, "Input HDF5 file");
+
+  CLI11_PARSE(app, argc, argv);
+
+  dunedaq::hdf5libs::HDF5RawDataFile decoder(filename);
+
+  auto trigger_record_numbers = decoder.get_all_trigger_record_numbers();
+  
+  // Populate the map with the TRHs and DS fragments
```
philiprodrigues, line 33: Oops I forgot to remove this now-obsolete comment. Could you remove it please?

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -108,6 +154,18 @@ TimingTriggerCandidateMaker::receive_hsievent(ipm::Receiver::Response message)
   auto data = serialization::deserialize<dfmessages::HSIEvent>(message.data);
   ++m_tsd_received_count;
   
+  if (m_hsi_passthrough == true){
```
philiprodrigues, line 157: If you just want to check that no bits beyond the low byte are set, you could do it more straightforwardly with, eg: `if (data.signal_map & 0xffffff00 != 0) { // high bits are set }`. I think that would save you the work making `bitset`s

plugins/TimingTriggerCandidateMaker.hpp
```diff
@@ -55,6 +55,17 @@ class TimingTriggerCandidateMaker : public dunedaq::appfwk::DAQModule
 
   std::string m_hsievent_receive_connection;
 
+  // HSI Passthrough changes
+  std::atomic<bool> m_hsi_passthrough;
+  int hsi_pt_before;
```
philiprodrigues, line 60: DAQ coding style is to prefix member variables in classes with `m_`, so `m_hsi_pt_before`, `m_hsi_pt_after`, `m_trigger_bitmask` and `m_low_high_bits`

philiprodrigues: Style guide is here: https://github.com/DUNE-DAQ/styleguide

plugins/TimingTriggerCandidateMaker.hpp
```diff
@@ -55,6 +55,17 @@ class TimingTriggerCandidateMaker : public dunedaq::appfwk::DAQModule
...
+  int hsi_pt_after;
+  std::bitset<16> trigger_bitmask;
+  std::vector< std::bitset<8> > LowHighBits;
+  std::bitset<16> MakeBitmask16(uint16_t signal_map);
```
philiprodrigues, line 64: DAQ style is `snake_case` for function names, so `make_bitmask16`, `make_bitmask8`, `split_bits` and `any_bit_set` (but this comment is moot if you use the alternative method for checking high bits that I suggested above)

philiprodrigues (changes requested):
Logic looks good. A couple of style changes suggested. I've done some tests and they work fine

plugins/ModuleLevelTrigger.cpp
```diff
@@ -148,10 +149,22 @@ ModuleLevelTrigger::create_decision(const triggeralgs::TriggerCandidate& tc)
...
+  if (m_hsi_passthrough == true){
+    if (tc.type == triggeralgs::TriggerCandidate::Type::kTiming){
+      decision.trigger_type = tc.detid;
```
philiprodrigues, line 157: Kurt suggests masking off the upper bits of `tc.detid`, in case they're set, so they don't leak into the wrong part of the trigger type. Let's do that, so `decision.trigger_type = tc.detid & 0xff;`

plugins/ModuleLevelTrigger.cpp
MRiganSUSX: This comment was there before my changes, specifically before the line:
decision.trigger_type = 1; // m_trigger_type;
This line is still there if the passthrough option is not set, so maybe we still want to change the assignment in the future (ie keep the comment, just move it to just before that line)? 
I have deleted it for now, but let me know what's better.

test/apps/print_trigger_type.cxx
MRiganSUSX: removed

plugins/TimingTriggerCandidateMaker.cpp
MRiganSUSX: I've added this.
Keep in mind there is a catch logic associated with this:
catch (SignalTypeError& e) {
    m_tc_sig_type_err_count++;
    ers::error(e);
    return;
which now won't get accessed? so the counter here won't count properly.

MRiganSUSX: If you'd like we can move the counter incrementation right after the ers::error( SignalTypeError ) call  and remove the catch completely, let me know.

plugins/TimingTriggerCandidateMaker.cpp
MRiganSUSX: Added this, which means I removed most of the nice functions I had fun creating, which are now obsolete :)

plugins/TimingTriggerCandidateMaker.hpp
MRiganSUSX: understood, renamed.

plugins/TimingTriggerCandidateMaker.hpp
MRiganSUSX: yup.

plugins/ModuleLevelTrigger.cpp
MRiganSUSX: added

plugins/ModuleLevelTrigger.cpp
philiprodrigues: I think it's fine for now, thanks

plugins/TimingTriggerCandidateMaker.cpp
philiprodrigues: Oh, my mistake. You were right the first time. I'll put it back how it was

philiprodrigues (approved):

Issue #94, opened 2022-03-18
Request to add run number to TPSet structure


bieryAtFnal:
...so that we can check for stale instances when one run ends and another one begins.

I originally noticed such a stop/start issue with TPSet in the TPStreamWriter in dfmodules.

PR #101, opened 2022-03-29
Add run number field to `Set<T>`

philiprodrigues:
This PR adds a `run_number` field to `Set<T>` (which adds it to `TPSet`s), as requested in #94.


bieryAtFnal (approved):
LGTM

Issue #96, opened 2022-03-21
Add collection/induction filter module for VD coldbox TPSets


philiprodrigues:
In the VD coldbox, the channel mapping is complicated and leads to collection and induction channels all being mixed up in the TPSets that are sent from readout. The easiest way to fix this is probably to have a module in trigger that reads in the TPSets and removes collection or induction TPs from them before passing them on to the next stage. Alternatively, the algorithm could do it, but `triggeralgs` doesn't depend on `detchannelmaps`

PR #103, opened 2022-04-04
Add TPFilter module

philiprodrigues:
This PR addresses #96 by adding `TPFilter`, a DAQModule which takes in `TPSet`s and modifies them to contain only `TriggerPrimitive`s from collection or induction channels (depending on configuration). The motivation for this module is the VD coldbox, where collection and induction channels are all mixed up, but the current trigger algorithms want to work on only collection channels.

### DIFF
--- a/plugins/TPChannelFilter.cpp
+++ b/plugins/TPChannelFilter.cpp
@@ -1,0 +1,119 @@
+/**
+ * @file TPChannelFilter.cpp
+ *
+ * This is part of the DUNE DAQ Application Framework, copyright 2021.
+ * Licensing/copyright details are in the COPYING file that you should have
+ * received with this code.
+ */
+
+#include "TPChannelFilter.hpp"
+#include "triggeralgs/TriggerPrimitive.hpp"
+
+#include <string>
+
+namespace dunedaq {
+namespace trigger {
+TPChannelFilter::TPChannelFilter(const std::string& name)
+  : DAQModule(name)
+  , m_thread(std::bind(&TPChannelFilter::do_work, this, std::placeholders::_1))
+  , m_input_queue(nullptr)
+  , m_output_queue(nullptr)
+  , m_queue_timeout(100)
+{
+
+  register_command("conf", &TPChannelFilter::do_conf);
+  register_command("start", &TPChannelFilter::do_start);
+  register_command("stop", &TPChannelFilter::do_stop);
+  register_command("scrap", &TPChannelFilter::do_scrap);
+}
+
+void
+TPChannelFilter::init(const nlohmann::json& iniobj)
+{
+  try {
+    m_input_queue.reset(new source_t(appfwk::queue_inst(iniobj, "tpset_source")));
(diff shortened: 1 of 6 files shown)


plugins/TPChannelFilter.cpp
```diff
@@ -0,0 +1,126 @@
...
+TPChannelFilter::do_scrap(const nlohmann::json&)
+{}
+
+bool
+TPChannelFilter::channel_should_be_removed(int channel) const
+{
+  // The plane numbering convention is found in detchannelmaps/plugins/VDColdboxChannelMap.cpp and is:
+  // U = 0, Y = 1, Z = 2
+  uint plane = m_channel_map->get_plane_from_offline_channel(channel);
+  return ((plane == 0 || plane == 1) && m_conf.keep_induction) || (plane == 2 && m_conf.keep_collection);
```
eflumerf, line 76: I'm confused by this logic. In my testing, I ran with both keep_induction and keep_collection false, but no TPs were removed. My supposition is that if the plane is not 0, 1, or 2, then this will return false and the channel will not be removed. The code might be more clearly expressed as follows:
```C++
// U = 0, Y = 1, Z =2
// Check for collection
if(plane == 0 || plane == 1) return !m_conf.keep_collection
// Check for induction
if(plane == 2) return !m_conf.keep_induction;
// Unknown plane?!
TLOG() << "Encountered unexpected plane " << plane << " from channel " << channel << ", check channel map?";
return false;
```

python/trigger/test_channel_filter/toplevel.py
```diff
@@ -0,0 +1,56 @@
...
+
+console = Console()
+
+import click
+
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('-s', '--slowdown-factor', default=1.0)
+@click.option('-f', '--input-file', type=click.Path(exists=True, dir_okay=False), multiple=True)
+@click.option('--keep-collection', is_flag=True, default=True, show_default=True, help="Keep collection TPs")
+@click.option('--keep-induction', is_flag=True, default=True, show_default=True, help="Keep induction TPs")
```
eflumerf, line 19:
```py
@click.option('--keep-collection/--discard-collection', default=True, show_default=True, help="Keep collection TPs")
@click.option('--keep-induction/--discard-induction', default=True, show_default=True, help="Keep induction TPs")
```

plugins/TPChannelFilter.cpp
philiprodrigues: You're right - I had a sign error. After a couple of attempts I think I've finally got it right. I also added handling for plane number 9999, which indicates channels in the VD coldbox that aren't actually connected to a wire - we unconditionally drop those.

python/trigger/test_channel_filter/toplevel.py
philiprodrigues: Nice! I didn't know about this feature. I've added this.

eflumerf (approved):
LGTM. Options appear to work correctly and the module now removes TPs as expected.

PR #104, opened 2022-04-06
Add streamed_TPs_to_text app

philiprodrigues:
This application reads the streamed TPs from an HDF5 file and writes them into the text format used by trigger.


bieryAtFnal:
Target Release?  (I set the Project equal to "DUNE-DAQ tracking", so attributes like Target Release can now be set.
Thanks

bieryAtFnal (approved):
I've run the new app on a file of streamed TPs and seen that a text file of approximately the same size is created.  The contents of the output file look reasonable, but I didn't use any Trigger tools to read it in.

CharlieBatchelor (approved):
I have successfully used the app on a `tpstream_*.hdf5` file to generate a list of TPs in a text file using a command like the one below. This resulting .txt file can be passed to the fakeTPChain, or the more recent `replay_tps_to_chain`, where I have managed to regenerate TAs and TCs.

`streamed_TPs_to_test -i tpstream_run000111_0000_tpwriter_20220421T151845.hdf5 -o tps.txt`

PR #110, opened 2022-04-21
Buffer TPs, TAs and TCs using the readoutlibs buffer implementation

### LINKED ISSUE #97: Read out TAs and TCs in response to data requests
`TPSetBufferCreator` already reads out TPs in response to data requests. We want the same logic for TAs and TCs

### LINKED ISSUE #109: Use latency buffer/fragment response code from readout
We currently have our own buffering/readout response code in `trigger`. If we switch to using the implementation from `readoutlibs`, we'll get consistent behaviour for free, and reduce the code maintenance required from `trigger`. Some preliminary work towards this end is in the philiprodrigues/readoutlibs-buffer branch.

philiprodrigues:
This PR changes the trigger object buffer implementation to use the one from readoutlibs, and adds buffering for TAs and TCs, in addition to TPs. It relies on the following two PRs:

https://github.com/DUNE-DAQ/detdataformats/pull/19
https://github.com/DUNE-DAQ/readoutlibs/pull/16


CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ find_package(daqdataformats REQUIRED)
 find_package(detdataformats REQUIRED)
 find_package(detchannelmaps REQUIRED)
 find_package(hdf5libs REQUIRED)
+find_package(readoutlibs REQUIRED)
```
eflumerf, line 22: Is there an entry in cmake/triggerConfig.cmake.in?

philiprodrigues: Oops, I forgot that. Thanks for catching it. Fixed now

philiprodrigues:
Some more description of what to expect, motivated by questions from @bieryAtFnal :

* TCs are generated inside the trigger app for real and fake HSI triggers, so there should be at least one TC read out for every trigger record. (I need to check this actually works for HSI triggers).
* TP-based prescale triggers should also produce TAs and TCs that get buffered and read out. The number of TAs and TCs that end up in the trigger record will depend on the readout window (set by the TC) and the rate of TAs/TCs. There should be at least one of each, if I'm thinking this through correctly.
* https://github.com/DUNE-DAQ/daqconf/pull/36 describes how the region and element IDs are generated for storage in the HDF5 file

PR #113, opened 2022-04-22
Fix copy-paste error setting opmon variables

philiprodrigues:
I made a copy-paste error in setting the "paused" time and "dead" time opmon variables. Fixed by this commit.


MRiganSUSX (approved):
this seems correct to me

PR #115, opened 2022-04-27
Set `origin` and `seqno` in TPSets emitted by FakeTPCreatorHeartbeatMaker

philiprodrigues:
This PR adds a correct setting of the `origin` and `seqno` fields in the TPSets emitted by `FakeTPCreatorHeartbeatMaker`


CharlieBatchelor (approved):
Tried this with my build of 2.11 and looks good. See good TPSets at triggeralgs.

PR #116, opened 2022-04-27
Correctly handle the outputting of heartbeat Sets in GenericMaker

philiprodrigues:
We were previously sending out heartbeat Set<T>s as soon as we received
them, but that was wrong, because the output buffer was potentially
holding Ts with times earlier than the heartbeat's time. Those Ts were
then output later than the heartbeat, breaking the ordering guarantee.

So now, TimeSliceOutputBuffer holds _two_ buffers: the buffer of
Ts (which it had before) and a new buffer of heartbeat
Set<T>s. Heartbeat Set<T>s are added via the buffer_heartbeat()
function. Adding a heartbeat to the buffer advances the latest seen
time, which is used to determine what data is ready to be sent
out. When flush() is called, the return (by reference) value is the
next heartbeat Set<T> if it sits at the start of the current time
window; otherwise it is a payload Set<T> with the Ts that fall in the
current window.


philiprodrigues:
To test, make a workdir with `daqconf` @ `philiprodrigues/test-generic-maker-heartbeats`, and trigger (initially @ develop, then @ this branch).

A configuration that will show the problem (without this PR):

```daqconf_multiru_gen --host-ru localhost --host-ru localhost --region-id 0 --region-id 1 --host-df localhost -d frames.bin -o . -s 1  --number-of-data-producers 2 -t 0 --enable-software-tpg --trigger-activity-config 'dict(prescale=500)' --trigger-candidate-config 'dict(prescale=20)' json```

Run nanorc on that configuration, and the `trigger` log will contain warnings like:

```2022-May-04 12:51:18,274 WARNING [bool dunedaq::trigger::TriggerZipper<TSET>::proc_one() [with TSET = dunedaq::trigger::Set<triggeralgs::TriggerActivity>] at /home/phrodrig/dunedaq-v2.10.0-nightlies/buffer-N22-04-21/sourcecode/trigger/plugins/TriggerZipper.hpp:209] Tardy input set from region 0 element 0. Set start time 81979124071830000 but last sent time 163958248156780000 DAQModule: tazipper```

With this PR, those warnings should be gone.

PR #117, opened 2022-05-04
Delete some obsolete "fake" test plugins and python confgens

philiprodrigues:
The fake test plugins were created for testing the trigger modules
when there wasn't a "real" version of the other end for them to talk
to. Eg, FakeDataFlow fakes the part of the data flow subsystem that
the MLT talks to, but we now have a real DFO to communicate with, so
FakeDataFlow is obsolete.

Removing these should reduce the amount of code we have to
maintain. (The immediate motivation for doing this now is to reduce
the amount of code that has to be updated for the unified
communications API changes).


bieryAtFnal (approved):
LGTM.  Everything built fine, and the typical integrationtests from the dfmodules repo all worked as expected.  The TriggerZipper unit test failed, but it was failing before this change, so that seems unrelated.

PR #118, opened 2022-05-09
Fix a case where TPM could produce TPSets with kUnknown type

philiprodrigues:


CharlieBatchelor (approved):
Used this fix to produce TRs with HMA run on ColdBox frames. Incoming TPSets look good.

PR #126, opened 2022-05-20
Use DefaultSkipListRequestHandler from readoutlibs

philiprodrigues:
This is needed to handle out-of-time-order objects in the latency buffer. Requires https://github.com/DUNE-DAQ/readoutlibs/pull/24 .

This fixes errors like:

```GeoID[type: DataSelection, region: 20000, element: 0] Trigger Matching result with empty fragment: TS match result on link 0: Trigger number=1 Oldest stored TS=79554162196262608 Start of window TS=79554162089472543 End of window TS=79554162089472893 Estimated newest stored TS=79554162089472543 Requestor=fragments_to_dataflow1```

which occur when TCs arrive out-of-order.

### DIFF
--- a/plugins/TABuffer.hpp
+++ b/plugins/TABuffer.hpp
@@ -12,8 +12,8 @@
 #include "daqdataformats/Fragment.hpp"
 #include "iomanager/Receiver.hpp"
 #include "readoutlibs/FrameErrorRegistry.hpp"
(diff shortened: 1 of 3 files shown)


bieryAtFnal (approved):
I've confirmed that it fixes some of the 3ru_3df_multirun_test SWTPG error messages.

PR #130, opened 2022-06-01
Philiprodrigues/use receive noexcept

bieryAtFnal:
As Phil requested, we're filing a PR to merge these changes once we have had a chance to test them.  We *have* tested them, and we see that the Trigger Inhibit messages are now gone from the SWTPG part of the 3ru_1df_multirun_test integration test.


eflumerf (approved):
My tests showed improvement, specifically that the trigger inhibit messages went away and the WIB system of the 3ru_1df integration test passed successfully.

PR #132, opened 2022-06-09
Files to start a michel electron trigger algorithm.

CharlieBatchelor:
Some additional plugin files to get started with the Michel electron trigger algorithm.


dingp:
Converted the PR to draft since we do not want this to be merged to `develop` before we freeze `dunedaq-v3.0.0` release.

PR #135, opened 2022-06-10
Added information to the debug message when a DataRequest is received…

bieryAtFnal:
…, to aid in debugging situations like long-window-readout in which multiple DRs are received per trigger.

I found this useful when debugging Issue 134, and my sense is that the additional debugging information may again be useful in the future.

This is not intended for v3.0.0, so I will set the Target Release to v3.1.0.


philiprodrigues (approved):
LGTM

PR #145, opened 2022-07-13
Added/changed some debug messages in the MLT

bieryAtFnal:
@mroda88 , will the addition of the run number in the Triggers PAUSED/RESUMED messages break anything in the Grafana displays?  I purposely only added text to the end of those messages in case some monitoring process is searching for "Triggers PAUSED/RESUMED"...


mroda88 (approved):
It's good for me. 
Tomorrow in my large scale test I'll use this repo too. 

I'm sure it will be alright.

mroda88 (approved):
Tested, all is good

PR #149, opened 2022-07-15
Add stop transition functionality

jcfreeman2:
Pierre pointed out we need to add functionality supporting the new approach to stop transition


jcfreeman2:
Force merge since Pierre says he can't pull the branch himself.

PR #151, opened 2022-07-27
Update `replay_tp_to_chain`

philiprodrigues:
This PR updates the `replay_tp_to_chain` application that reads TPs from text file and replays them through the trigger system, so that it works with the latest trigger app from daqconf, and the latest daqconf style for connections etc. Requires https://github.com/DUNE-DAQ/daqconf/pull/153


CharlieBatchelor (approved):
Tested and works as expected.

PR #156, opened 2022-09-20
Get TP replay app working with new source ID changes and new trigger type

CharlieBatchelor:
The changes made here are to fix the way we generate `df_conf` in a way that is suitable for passing to `get_dfo_app()` with the new source ID code.

Some further fixes were required after Pierre’s changes to `daqconf` such as the boot configuration generation needed for creating app command datas.


roland-sipos (approved):
LGTM

eflumerf (approved):
LGTM

PR #160, opened 2022-09-27
Option to provide TC types to be ignored in the MLT

MRiganSUSX:
passed via the daqcong config json
loaded in MLT
incoming TC type check against the vector
if to be ignored: skips the buffering logic and does not contribute to TD decision making

also added to opmon and run summary in MLT


philiprodrigues (approved):
LGTM

wesketchum:
@MRiganSUSX I'm guessing we may want to test that as part of the integration week? If so, let's wait to merge this in / create a PR against a patch/v3.2.x branch.

PR #164, opened 2022-10-20
EHN1: Option to provide TC types to be ignored in the MLT

wesketchum:
Repeat of #160 , but going to integration week patch branch.


MRiganSUSX:
Sorry, is this not getting merged to patch for testing? Cheers.

PR #165, opened 2022-11-03
EHN1: Option to provide TC types to be ignored in the MLT

wesketchum:
Repeat of #160 , but to patch branch.

### DIFF
--- a/schema/trigger/moduleleveltrigger.jsonnet
+++ b/schema/trigger/moduleleveltrigger.jsonnet
@@ -10,6 +10,8 @@ local types = {
   td_out_of_timeout_b : s.boolean("td_out_of_timeout_b"),
   candidate_type_t : s.number("candidate_type_t", "u4", doc="Candidate type"),
   time_t : s.number("time_t", "i8", doc="Time"),
+  tc_type : s.number("tc_type", "i4", doc="TC type"),
+  tc_types : s.sequence("tc_types", self.tc_type, doc="List of TC types"),
 
   sourceid : s.record("SourceID", [
       s.field("element", self.element_id, doc="" ),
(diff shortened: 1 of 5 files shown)


schema/trigger/moduleleveltrigger.jsonnet
```diff
@@ -27,6 +29,7 @@ local types = {
       s.field("td_out_of_timeout", self.td_out_of_timeout_b, doc="Option to drop TD if TC comes out of timeout window"),
       s.field("buffer_timeout", self.time_t, 100, doc="Buffering timeout [ms] for new TCs"),
       s.field("td_readout_limit", self.time_t, 1000, doc="Time limit [ms] for the length of TD readout window"),
+      s.field("ignore_tc", self.tc_types, [], doc="List of TC types to be ignored"),
```
wesketchum, line 32: This is `ignore_tc`, while daqconf uses an `mlt_ignore_tc`.

wesketchum (commented):
Not working at the moment with the daqconf change too.

schema/trigger/moduleleveltrigger.jsonnet
MRiganSUSX: `ignore_tc` is linked to `mlt_ignore_tc` in daqconf, file python/daqconf/apps/trigger_gen.py of PR 259 as:
ignore_tc=MLT_IGNORE_TC,

this is not an issue and we were able to build and generate configs with this with no issues. What is the exact error you are seeing? Cheers.

wesketchum:
OK, we've tested this and I believe we can say it works at EHN1. @MRiganSUSX @CharlieBatchelor can you review/approve (technically I made this PR to the patch branch) so we can merge in?

CharlieBatchelor (approved):
The changes have been tested in a SWTPG run at EHN1. The expected output from `trigger` point of view was shown in the logs for that run. Additionally, these didn't interfere with other `trigger` related tests that took place in the same run. Looks okay to be merged to the patch branch.

PR #169, opened 2022-11-30
Revert changes to TPChannelFilter

CharlieBatchelor:
Some changes were made to the file `TPChannelFilter.cpp` during a SW TPG run of the integration week. These should not have made their way through to the `patch/v3.2.x` branch of trigger. This is to revert that change and provides a test fix to the over-production of TRs in the SW TPG integration tests.


wesketchum:
Looks good, and tested at EHN1 (and it fixes 3ru_3df integration test).

wesketchum (approved):

PR #172, opened 2022-12-08
Revert changes to TPChannelFilter for integration week tests.

CharlieBatchelor:
A temporary change in the code made its way to `develop` accidentally during integration week. This one liner fix reverts that change and has been shown to produce good integration test results for the SW TPG issues @bieryAtFnal raised.


CharlieBatchelor:
@bieryAtFnal - as discussed, the equivalent merge was done via your branch.

PR #173, opened 2022-12-08
Merge most recent change on patch/v3.2.x to develop

bieryAtFnal:


CharlieBatchelor (approved):
The changes reflect those I made for `patch/v3.2.x` to revert the way in which induction channels are ignored by the `TPChannelFilter`. Approved.

PR #177, opened 2023-01-17
fix the issue 176

plasorak:
problem in endpoint specification
defaults use connectivity service in daqconf


CharlieBatchelor (approved):
I'm able to successfully generate a `replay_tp_app` configuration and run to completion with the above changes. Approving and merging.

PR #179, opened 2023-01-26
Fixes and enhancements for patch/v3.2.x VDCB tests

CharlieBatchelor:
Tolerate incompleteness in a controlled way for the VDCB runs. Requires another PR in `daqconf`

PR #180, opened 2023-01-26
Implement Brett's zipper fix for VDCB runs

CharlieBatchelor:
The max tolerance mode was not being observed at EHN1 VDCB runs, and Brett has created a fix to ensure this is now working as expected. That fix is also implemented here where we can test in the upcoming VDCB runs.


ArturSztuc (approved):
Looks good, nothing suspicious in `3ru_3df_multirun_test.py` integration test either.

CharlieBatchelor:
Merged these changes to the completeness tolerance branch, closing for #179.

PR #182, opened 2023-02-02
Updating CMakeLists.txt

CharlieBatchelor:


wesketchum (approved):
increase by two due to other tag, but better safe than sorry.

PR #183, opened 2023-02-02
Mrigansusx/mlt readout map

MRiganSUSX:
Addition of readout map to MLT, allows configurable readout windows per TC type.
Updated logic for ignored TC types in MLT: now added to list of contributing TCs in TD (so can be later used in ROI, coincidences), but does not extend the readout window or expiration.


MRiganSUSX:
This is now superseded by #194

PR #189, opened 2023-03-01
Eflumerf/check run number

eflumerf:
It appears that this change is still valid...


eflumerf:
Not sure if this is still needed, or who the correct reviewer may be.

eflumerf:
After talking with trigger folks, this should probably be reimplmented in the TPChannelFilter.

PR #191, opened 2023-03-06
update connections to always have local names

glehmannmiotto:
change to setup connection endpoint at init, to be more uniform with other modules and the use of the connectivity service.


bieryAtFnal (approved):
LGTM.
I ran several integtests, looked through the code differences, and confirmed that there aren't get_iom_receiver/sender calls outside of the init() method (except in one place, and in that one place, the connection name was stored in a data member in the init() method).

PR #198, opened 2023-03-27
Fix readout map configuration.

CharlieBatchelor:
This fixes the recent issue with nightly N23-03-26, where the WIB Fragments size are larger than expected due to the new MLT configurable readout window sizes per TC type. It is coupled to the `daqconf` [#309](https://github.com/DUNE-DAQ/daqconf/pull/309).


CharlieBatchelor:
I checked these changes work with the `minimal_system_quick_test.py` and it runs to completion without error. Will merge immediately to fix the issue.

PR #199, opened 2023-03-27
Mrigansusx/mlt readout map fix

MRiganSUSX:
A fix for the MLT readout map.
It makes the readout map optional, and it is false by default. 
If the map is to be used, the readout windows are passed from daqconf configuration.
If the map is not to be used, MLT uses TC.time_start and TC.time_end, as was the case before the use of the map.


bieryAtFnal (approved):
LGTM, in terms of the behavior of dfmodules/integtests.

PR #218, opened 2023-04-28
Mrigansusx/mlt merging flag

MRiganSUSX:
An addition of configurable flag that can enable/disable merging of overlapping TCs when forming TDs in MLT.


bieryAtFnal (approved):
LGTM.  I ran a test which had readout windows that were long enough to generate the 'overlapping' errors with the old code, and these errors went away with the new code and the mlt_merge_overlapping_tcs parameter set to False.

Issue #220, opened 2023-06-23
MLT: Coincidence triggering


MRiganSUSX:
The trigger group deliverables for dunedaq-v4.1.0 release are discussed here: 
[trigger deliverables for v4.1.0](https://indico.fnal.gov/event/59398/contributions/264506/attachments/165845/220470/ppds_deliverables_v4.1b.pdf)

The main trigger deliverable is the ability to have coincidence triggering: 
- being able to specify TC types to trigger on (configurable)
- have the ability to trigger on the coincidence of specified TC types

The design and implementation are discussed here:
[MLT: Coincidence triggering](https://indico.fnal.gov/event/60211/contributions/269136/attachments/167981/224786/report.pdf)

Brief overview:
- coincidence is described as overlapping readout windows of TCs (using the already present MLT merging functionality)
- using bitset as representation of bitwords (describing whether to trigger on particular TC, or coincidence)
- configuration done via daqconf:
     - mlt_use_bitwords (flag): whether to use bitwords when triggering
     - mlt_trigger_bitwords (sequence of sequences of strings): the object holding the configuration of bitwords, in names
     - mlt_trigger_bitwords_map (dictionary): the object holding the TC type map (does not need to be defined, kept in trgdataformats)

Example config:
#####
    "mlt_use_bitwords": true,
    "mlt_trigger_bitwords": [["kSupernova"],["kHorizontalMuon"],["kTPCLowE", "kTiming"]],
#####
- this represents 3 trigger bitwords. We would trigger on ANY supernova TC, ANY horizontal muon TC, and a coincidence of TPCLowW and Timing TCs.

MLT reports on triggered/not triggered objects:
#####
Run 111: Received 1048 TCs.
Sent 38 TDs consisting of 125 TCs.
12 TDs (38 TCs) were created during pause, and 8 TDs (25 TCs) were inhibited.
0 TDs (0 TCs) were dropped. 6 TDs (10 TCs) were cleared.
Not triggered (failed bitword check) on 742 TDs consisting of 850 TCs.
#####

Required branches:
- [mrigansusx/mlt_coinc in trigger](https://github.com/DUNE-DAQ/trigger/tree/mrigansusx/mlt_coinc)
- [mrigansusx/mlt_coinc in daqconf](https://github.com/DUNE-DAQ/daqconf/tree/mrigansusx/mlt_coinc)
[...]

PR #221, opened 2023-06-27
MLT: Coincidence trigger (trigger deliverable for v4.1.0)

MRiganSUSX:
This is the implementation of coincidence trigger as discussed in
[Trigger: Issue 220](https://github.com/DUNE-DAQ/trigger/issues/220)


plugins/ModuleLevelTrigger.cpp
```diff
@@ -394,49 +419,63 @@ ModuleLevelTrigger::send_trigger_decisions()
...
-  TLOG_DEBUG(3) << "Override?: " << override_flag;
-  if ((!m_paused.load() && !m_dfo_is_busy.load()) || override_flag) {
+
+  if (m_use_bitwords) {
+    // Check trigger bitwords
+    m_TD_bitword = get_TD_bitword(pending_td);
+    m_bitword_check = check_trigger_bitwords();
+  }
+
+  if ((!m_use_bitwords) || (m_bitword_check)) {
```
ArturSztuc, line 429: This is a big if-else. Could easily convert into small one by reversing and return, like so:

```
if( !((!m_use_bitwords) || (m_bitword_check)){
    m_td_not_triggered_count++;
    m_td_not_triggered_tc_count += pending_td.contributing_tcs.size();
    return;
}
```

And the rest of the code in that function (that was above, and now below) here, completely outside large if statement

plugins/ModuleLevelTrigger.hpp
```diff
@@ -137,11 +137,22 @@ class ModuleLevelTrigger : public dunedaq::appfwk::DAQModule
   int m_earliest_tc_index;
   int get_earliest_tc_index(const PendingTD& pending_td);
 
+  // Bitwords logic
+  bool m_use_bitwords;
+  nlohmann::json m_trigger_bitwords_json;
+  bool m_bitword_check;
+  std::bitset<16> m_TD_bitword;
+  std::vector<std::bitset<16>> m_trigger_bitwords;
+  std::bitset<16> get_TD_bitword(const PendingTD& ready_td);
```
ArturSztuc, line 146: I'm wondering if the number 16 could be placed in some more more global variable so it's not hard-coded here (and many other places). Probably not configurable?

ArturSztuc (approved):
Overall looks good. I made a couple of comments/recommendations, but they are minor and not significant -- so I don't want them to stop this PR.

plugins/ModuleLevelTrigger.cpp
MRiganSUSX: Thanks for the comment. That would probably look nicer but does not change the logic. I will keep that in mind for future dev but will skip it for now.

plugins/ModuleLevelTrigger.hpp
MRiganSUSX: As we discussed offline, this is not straightforward and would likely need dedicated spot somewhere outside daqconf. Let's keep it as is for now and revisit at later point.

PR #224, opened 2023-07-10
fix for the stop command logic in TP Channel Filter

MRiganSUSX:
An issue was discovered where the trigger app would not stop in time - especially reproducible when running with multiple readouts. The issue was tracked back to the TPChannelFilter. The logic for checking the stop command was updated and did fix the issue for a particular test. (more testing needed)


bieryAtFnal (approved):
LGTM, and I've run a subset of the integtests to confirm that existing functionality works.

PR #240, opened 2023-09-20
fixes for random and custom TC maker plugins

MRiganSUSX:
A fix to make RandomTriggerCandidateMaker functional, including all configuration parameters. 

RTCM is a standalone module that inserts TCs straight into the MLT. 
These TCs can be created at a configurable frequency. 
The are two methods when creating these: a flat distribution (kUniform) and a random Poisson distribution (kPoisson). 
The time estimation can be done either via system clock or using the TimeSync messages. 
More details are available here: [slides](https://indico.fnal.gov/event/61371/contributions/275588/attachments/171005/230370/RTCM.pdf)

Example configuration for testing:
 ```
"trigger": { 

    "use_random_maker": true,
    "rtcm_timestamp_method": "kSystemClock",
    "rtcm_time_distribution": "kUniform",
    "rtcm_trigger_interval_ticks": 32000000

  }
```
  
  Minimal configuration for testing (that uses defaults):
 ```
"trigger": {
    "use_random_maker": true,
...
  }
```

### DIFF
--- a/plugins/CustomTriggerCandidateMaker.cpp
+++ b/plugins/CustomTriggerCandidateMaker.cpp
@@ -58,7 +58,7 @@ CustomTriggerCandidateMaker::CustomTriggerCandidateMaker(const std::string& name
 void
 CustomTriggerCandidateMaker::init(const nlohmann::json& obj)
 {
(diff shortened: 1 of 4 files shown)


bieryAtFnal (approved):
I've verified that a representative subset of existing integtests continue to succeed with these changes.  My vote is to go ahead and merge these changes and work on new integtest(s) that make use of the Random TC maker as a separate step.

PR #244, opened 2023-10-31
Possible readout window change in MLT when use_roi_readout is true 

bieryAtFnal:
In ModuleLevelTrigger::roi_readout_make_requests(), this change bases the readout window on trigger time plus/minus the configured time window...

Not sure if this change is reasonable, but without it, we get no data fragments from the DataLinkHandlers in the Readout Apps or TPBuffers in the Trigger App.  To be discussed at the next Release Coordination meeting...


MRiganSUSX:
This is a bug and the fix above is the intended behaviour.

MRiganSUSX (approved):

PR #247, opened 2023-11-10
Updated tlogs in MLT

MRiganSUSX:
Simple improvements to TLOG_DEBUG values + additional TLOGS for TD readout windows, and functions where readout windows are being set.

### DIFF
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -222,7 +222,7 @@ ModuleLevelTrigger::do_pause(const nlohmann::json& /*pauseobj*/)
   m_livetime_counter->set_state(LivetimeCounter::State::kPaused);
   TLOG() << "******* Triggers PAUSED! in run " << m_run_number << " *********";
   ers::info(TriggerPaused(ERS_HERE));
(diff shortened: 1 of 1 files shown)


bieryAtFnal (approved):
LGTM.
Enabled TLVLs 3, 5, and 10 and looked at the messages.
Verified that a few integtests work fine.
(I re-merged the develop branch into this branch in order to test these changes with a recent nightly build.)

ArturSztuc (approved):
lgtm

MRiganSUSX:
Thanks both for the review. I will merge in a bit.
One more question, probably for @bieryAtFnal, is there any normalized way of doing these TLOG_DEBUG levels? I've seen in some other portions of code things like:
[TLOG_DEBUG level enum example](https://github.com/DUNE-DAQ/dfmodules/blob/develop/plugins/DataWriter.cpp#L34C1-L41C3)
Thanks.

bieryAtFnal:
Hi @MRiganSUSX , yes, in some places, we use constants to define TRACE levels.  I don't think that we have a consistent strategy for doing that.  It is more of a convenience, to provide a string name that gives developers a sense of what the authors intended to provide with the various sets of TRACE messages.

PR #248, opened 2023-11-15
MLT readout map rework

MRiganSUSX:
Fix for custom readout map used in MLT to only apply custom readout windows to explicitly configured TC types.

### DIFF
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -333,7 +333,7 @@ ModuleLevelTrigger::send_trigger_decisions()
   while (m_running_flag) {
     std::optional<triggeralgs::TriggerCandidate> tc = m_candidate_input->try_receive(std::chrono::milliseconds(10));
     if (tc.has_value()) {
-      if (m_use_readout_map) {
+      if ( (m_use_readout_map) && (m_readout_window_map.count(tc->type)) ) {
         TLOG_DEBUG(1) << "Got TC of type " << static_cast<int>(tc->type) << ", timestamp " << tc->time_candidate
                       << ", start/end " << tc->time_start << "/" << tc->time_end << ", readout start/end "
                       << tc->time_candidate - m_readout_window_map[tc->type].first << "/"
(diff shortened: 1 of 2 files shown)


MRiganSUSX:
More details - with description, configurations, testing, and results - available here: 
[custom_readout_map_fix.pdf](https://github.com/DUNE-DAQ/trigger/files/13377208/custom_readout_map_fix.pdf)
Required: [daqconf PR412](https://github.com/DUNE-DAQ/daqconf/pull/412)

plugins/ModuleLevelTrigger.cpp
```diff
@@ -509,12 +509,12 @@ ModuleLevelTrigger::add_tc(const triggeralgs::TriggerCandidate& tc)
 
     for (std::vector<PendingTD>::iterator it = m_pending_tds.begin(); it != m_pending_tds.end();) {
       if (check_overlap(tc, *it)) {
-        if (m_use_readout_map) {
+	it->contributing_tcs.push_back(tc);
+        if ( (m_use_readout_map) && (m_readout_window_map.count(tc.type)) ){
```
ArturSztuc, line 513: I cannot select this all, but lines 511 - 536, that's a very, nested if statement,  makes it difficult to read and understand... normally something like this:

```
for(i = 0; i < n; ++i){
   if(this_true){
      if(that_als_true){
         50 lines of code
      }
   }
}
```

can be easily changed withs simple guard clauses:

```
for(i = 0; i < n; ++i){
   if(!this_true)
      continue;
   if(!that_also_true)
      continue;

   50 lines of code
}
```

Can read more about them on the refactoring guru page: https://refactoring.guru/replace-nested-conditional-with-guard-clauses

It's a comment though, PR still approved :+1:

ArturSztuc (approved):
LGME, although future guard clauses in MLT would be nice for readability.

bieryAtFnal (approved):
LGTM.

I've run tests in which I've separately controlled the readout window for two individual TC types in an emulated-data system in which I have three types of TCs being produced (timing, prescale, random).

As part of this, I made some enhancements to the `HDF5LIBS_TestDumpRecord` test application to print out the request window that was used for each SourceID in a TriggerRecord on disk.  I'll contribute those independently of this PR.

PR #251, opened 2023-11-27
Adding missing variables in TP streamer app

emanuele-villa:
streamed_TPs_to_text.cpp was missing the printout of the variables `algorithm`, `version` and `flag`. Added them for completeness and consistency with other tools.

This change should not affect anything else. Branch can be deleted during the merge.


emanuele-villa:
I was too naive.
Need to take care of the casting of custom defined data types such as Algorithm and Type when printing them out to file. Closing this request, will open a new one after checking carefully that there are no issues.

PR #254, opened 2023-12-04
Trigger Activity Factory Refactorization

aeoranday:
Refactors cetlibs dependent plugin structure of singleton factory design pattern for TA Maker creation.

TriggerActivityFactory contains registered TA Maker plugins and returns a singleton on request. This has a corresponding PR in https://github.com/DUNE-DAQ/triggeralgs that defines the structure of this factory.


ArturSztuc (approved):
LGTM

aeoranday:
If there are no objections, I will go ahead and merge here and in `triggeralgs` on Tuesday morning CEST.

MRiganSUSX:
Sorry, late to the party here.
Looks good, also support the direction of this change. Is a 'Trigger Candidate Factory' planned?

aeoranday:
Yes, I plan to do an abstract factory that both TA and TC can be derived from.

This at least showed the proof of concept before going too far into it.

PR #258, opened 2023-12-26
Removed unused trigger_interval_ticks parameter from customtriggercandidatemaker.jsonnet.

bieryAtFnal:
To avoid confusion...


MRiganSUSX (approved):
Thanks for finding this (an obsolete piece carried over from random TC maker, I think).
LGTM

Issue #250, opened 2023-11-24
MLT occasionally sends additional TriggerDecisions to the DFO after it completes the 'disable_triggers' transition


bieryAtFnal:
This should not happen.  The meaning of the 'disable_triggers' command/transition is to stop sending TriggerDecisions.

I suspect that this is caused by the MLT sending out 'pending' TDs at 'stop' time, and my sense is that it should instead do that at 'disable_triggers' time.  But, I'll leave the full diagnosis and fix to the experts.

Here are TRACE messages that provide some insight into what is happening...

Below, in reverse time order, are TRACE messages that show the MLT pausing triggers (lines 13552-13554); the DFO completing the “stop” transition (line 13549); the MLT flushing pending TDs, sending a TD to the DFO, and completing the “stop” transition (lines 13451-13456), and the DFO complaining about a TD with the wrong run number at the start of the next run (line 13412).

Recall that run control commands are broadcast to all DAQ applications without any deliberate ordering.  So, it is not surprising that the DFO could complete “stop” before the MLT does.  That is exactly why the “disable_triggers” command/transition was added (to ensure that the flow of TriggerDecisions are stopped before the DFO is told to stop the run).

  13412 11-23 20:17:46.058799  672746  673582  78          DataFlowOrchestrator ERR receive_trigger_decision: DataFlowOrchestrator encountered run number mismatch: recvd (101) != 102 from MLT for trigger_number 462
  13418 11-23 20:17:46.058168  672746  673061 120          DataFlowOrchestrator LOG add_callback_impl: Registering callback.
  13419 11-23 20:17:46.058004  672746  673061 120          DataFlowOrchestrator LOG add_callback_impl: Registering callback.
  13451 11-23 20:17:43.217897  672967  673077  32            ModuleLevelTrigger NFO do_stop: End of run 101
  13453 11-23 20:17:43.217178  672967  673077  32            ModuleLevelTrigger LOG do_stop: LivetimeCounter - total deadtime+paused: 2396
  13454 11-23 20:17:43.216914  672967  673335  44            ModuleLevelTrigger LOG send_trigger_decisions: Run 101: Received 494 TCs. Sent 462 TDs consisting of 462 TCs. 29 TDs (29 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped.
[...]

PR #262, opened 2024-01-18
Fix for MLT sending TDs after run stop (after 'disable_triggers' transition)

MRiganSUSX:
This PR fixes issues described in https://github.com/DUNE-DAQ/trigger/issues/250. 

TLDR: there was a (rare) bug where TDs could have ended in pending state (before their expiration was reached) at the run stop. The original idea was to send these (in order to not drop 'good' data). However, the original implementation attempted this too late, and the DFO could no longer process requests for these.
This PR fixes that by moving the flushing of pending TDs to an earlier stage.

Additionally, thanks to @bieryAtFnal, there is also an integration test now available that reproduces the original issue and also proves this PR fixes the issue.


bieryAtFnal (approved):
Looks good to me.  I verified that the new integtest that looks closely at this use case runs successfully, and that other representative integtests also run successfully.

PR #264, opened 2024-01-18
Fix for TC ignore check in MLT

MRiganSUSX:
Fix for logic error where a variable was assigned content after some logic comparison was done on it.
Thanks to @ArturSztuc for spotting that.

Small change.
Passes:

- basic tests from instructions page
- integration tests
- custom tests using the ignore functionality


ArturSztuc:
Perfect, thank you!

ArturSztuc (approved):
Looks perfect, thank you!

PR #267, opened 2024-01-19
Adding prescaling for ttcm

### LINKED ISSUE #265: Prescale option for Timing Trigger Candidate Maker

MRiganSUSX:


MRiganSUSX:
This PR (based on v4.2 as requested) adds the option to prescale TCs created in the Timing Trigger Candidate Maker.

The accompanying daqconf PR: https://github.com/DUNE-DAQ/daqconf/pull/420

The configuration parameter is:
`"ttcm_prescale": 5` <- any positive int here

----
Trigger logs from testing:
```
1: no pres conf
Executing conf -> ttcm
ttcm configured.

Run 1: Received 12 TCs. Sent 10 TDs consisting of 10 TCs. 2 TDs (2 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.

2: presc = 1 in conf
Executing conf -> ttcm
ttcm configured.

Run 2: Received 12 TCs. Sent 10 TDs consisting of 10 TCs. 2 TDs (2 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.

3: presc = 10 in conf
Executing conf -> ttcm
Runnig with prescale at: 10

Run 3: Received 1 TCs. Sent 1 TDs consisting of 1 TCs. 0 TDs (0 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.

4: presc = 5 in conf
Executing conf -> ttcm
Runnig with prescale at: 5

Run 4: Received 2 TCs. Sent 2 TDs consisting of 2 TCs. 0 TDs (0 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.
```

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -129,7 +134,13 @@ TimingTriggerCandidateMaker::receive_hsievent(dfmessages::HSIEvent& data)
   }
 
   ++m_tsd_received_count;
-  
+
+  if (m_prescale_flag) {
+    if (m_tsd_received_count % m_prescale != 0){
```
aeoranday, line 139: Is this the correct logic? Shouldn't it be `if (m_tsd_received_count % m_prescale == 0)`?

MRiganSUSX: Not in this case, if you look at what follows this if, it's `return`; meaning in the case of `==0` we would go on and continue processing the hsi event, creating TC, and sending it; while in the case of `!=0` we simply skip the whole processing and wait for another hsi event.

MRiganSUSX:
I again retested with (now after dev was merged):
- integration tests:
-- minimal_system_quick_test
-- fake_data_producer_test
-- 3ru_3df_multirun_test
- custom tests:
-- (1) not specifying anything in daqconf
-- (2) specifying the default in daqconf: ttcm_prescale = 1
-- (3) ttcm_prescale = 5
-- (4) ttcm_prescale = 10

The results are as expected, with TCs being scaled as configured:
```
1:
Executing conf -> ttcm
ttcm configured.
Run 5: Received 32 TCs. Sent 30 TDs consisting of 30 TCs. 2 TDs (2 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.

2:
Executing conf -> ttcm
ttcm configured.
Run 2: Received 32 TCs. Sent 30 TDs consisting of 30 TCs. 2 TDs (2 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.

3:
Executing conf -> ttcm
ttcm configured.
Running with prescale at: 5
Run 3: Received 6 TCs. Sent 6 TDs consisting of 6 TCs. 0 TDs (0 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.

4:
Executing conf -> ttcm
ttcm configured.
Running with prescale at: 10
Run 4: Received 3 TCs. Sent 3 TDs consisting of 3 TCs. 0 TDs (0 TCs) were created during pause, and 0 TDs (0 TCs) were inhibited. 0 TDs (0 TCs) were dropped. 0 TDs (0 TCs) were cleared.
```

aeoranday (approved):
LGTM. Thanks for clarifying that logic!

schema/trigger/timingtriggercandidatemaker.jsonnet
```diff
@@ -6,6 +6,7 @@ local types = {
 	time_t : s.number("time_t", "i8", doc="Time"),
 	signal_type_t : s.number("signal_type_t", "u4", doc="Signal type"),
 	hsi_tt_pt : s.boolean("hsi_tt_pt"),
+        count_t : s.number("count_t", "i8", doc="Counter"),
```
ArturSztuc: Does i8 mean 8-bit int? If so, then it allows for negative prescales (which wouldn't make sense, so u8 would be better), and only maximal prescale of 127 (so I would change it to uint16, which would allow for max of 65535, or even uint32).
Might be wrong how i8 is defined though, and how it works with the python -> cpp interface...

ArturSztuc (approved):
LGTM, but do you know how the i8 will work here (I wrote separate comment for the relevant chink of code).

schema/trigger/timingtriggercandidatemaker.jsonnet
MRiganSUSX: I'm not sure and I agree and share your scepticism about this. 
My reasoning was this:
in daqconf version of this schema (https://github.com/DUNE-DAQ/daqconf/blob/mrigan/ttcm_prescale/schema/daqconf/triggergen.jsonnet) there already existed a type for counting `types.count` which is defined here: https://github.com/DUNE-DAQ/daqconf/blob/mrigan/ttcm_prescale/schema/daqconf/types.jsonnet#L21C3-L21C75 as `i8`. I simply propagated this type to trigger as well.

MRiganSUSX: This type `count` is also used for many other variables that are definitely just supposed to be positive ints (like `trigger_window_before_ticks`)

MRiganSUSX: so options are:
- leave as is, since this is already widely used
- create a new type for this counting
- limit this specific type `count_t` to have a minimum of 1
Let me know

ArturSztuc: Ah, thank you for this. I can see [HERE](https://github.com/DUNE-DAQ/daqconf/blob/940cce5a98aaabf44f1589b94885369faa2a842b/schema/daqconf/types.jsonnet#L11-L16) that these numbers are defined as bytes, no bits. So i8 is actually standard 64-bit int, i4 would be 32-bit int etc. Makes sense, thank you! I still think u8 would be better than i8 since we cannot have negative pre-scale, but I don't think it matters that much.

ArturSztuc: You added more just before I pressed "Comment" :) Unless the other options are literally 5-min of work,  I would say leave as is for now (so your option 1)

MRiganSUSX: Thanks for the comments. I have now added a check that ensures the value is >=1. 
Otherwise, this error message is shown:
```ValueError: illegal count_t number 0 not greater than or equal 1```

Also confirmed previous tests still work as expected.

ArturSztuc: Thank you, in that case let's merge the PR :)

PR #268, opened 2024-01-24
TPM HDF5 Usage

aeoranday:
TriggerPrimitiveMaker currently uses a txt file to read TriggerPrimitives from. This change allows the TPM to use HDF5 files to read from with the requirement that the HDF5 file contains TimeSlices (implying the file is a tpstream).

This was tested using `generate_tpset_from_tpm.cxx`. This application focuses on the changes made to the TPM by copying the new code and running against an arbitrary tpstream file. This does not test the use of TPM in the replay application.


test/apps/generate_tpset_from_tpm.cxx
```diff
@@ -0,0 +1,107 @@
+/**
+ * @file generate_tpset_from_hdf5.cxx Read TP fragments from file and generate a TPSet vector.
```
ArturSztuc: This file is not called `generate_tpset_from_hdf5`

plugins/TriggerPrimitiveMaker.cpp
```diff
@@ -133,54 +133,69 @@ TriggerPrimitiveMaker::do_scrap(const nlohmann::json& /*args*/)
...
-
-        tpset.type = TPSet::Type::kPayload;
-
-        if (!tpset.objects.empty()) {
-          // We don't send empty TPSets, so there's no point creating them
-          tpsets.push_back(tpset);
+  for (std::string& fragment_path : fragment_paths) {
+    std::unique_ptr<daqdataformats::Fragment> frag = input_file->get_frag_ptr(fragment_path);
+    // Make sure this fragment is a TriggerPrimitive
+    if (frag->get_fragment_type() != daqdataformats::FragmentType::kTriggerPrimitive) continue;
```
ArturSztuc: I would also add something like:
```
      if (frag->get_element_id().subsystem != daqdataformats::SourceID::Subsystem::kTrigger)
           continue;
```
Otherwise there's a chance of getting non-trigger TPs? E.g. ones stored in the Readout rather than Trigger (artificially duplicating the amount of TPs present)

plugins/TriggerPrimitiveMaker.cpp
```diff
@@ -133,54 +133,69 @@ TriggerPrimitiveMaker::do_scrap(const nlohmann::json& /*args*/)
...
+
+          tpset.objects.clear();
         }
-        prev_tpset_number = current_tpset_number;
-
-        tpset.objects.clear();
+        tpset.objects.push_back(tp);
+      } else {
+        ers::warning(UnsortedTP(ERS_HERE, get_name(), tp.time_start));
       }
```
ArturSztuc: Instead, you could have just done 
```
(!(tp.time_start >= old_time_start))){
    ers::warning(UnsortedTP(ERS_HERE, get_name(), tp.time_start));
    continue;
}
````

at the beginning, and the rest of the code here would be outside of the if statement, closer to the right.
It's overall a bad coding practice to do `else continue;` at the end. Here's more info:
https://refactoring.guru/replace-nested-conditional-with-guard-clauses

File: plugins/TriggerPrimitiveMaker.cpp [hunk 4]
```diff
@@ -133,54 +133,69 @@ TriggerPrimitiveMaker::do_scrap(const nlohmann::json& /*args*/)
...
+      auto& tp = tp_array[i];
+      if (tp.time_start >= old_time_start) {
+        // NOLINTNEXTLINE(build/unsigned)
+        uint64_t current_tpset_number = (tp.time_start + m_conf.tpset_time_offset) / m_conf.tpset_time_width;
+        old_time_start = tp.time_start;
+
+        // If we crossed a time boundary, push the current TPSet and reset it
+        if (current_tpset_number > prev_tpset_number) {
+          tpset.start_time = prev_tpset_number * m_conf.tpset_time_width + m_conf.tpset_time_offset;
+          tpset.end_time = tpset.start_time + m_conf.tpset_time_width;
```
ArturSztuc: Is this how this is set in the readout? Does the config have a default value that's the same as readout's default value for TPSet creation?

test/apps/generate_tpset_from_tpm.cxx
```diff
@@ -0,0 +1,107 @@
...
+
+  // Read in the file and place the TPs in TPSets. TPSets have time
+  // boundaries ( n*tpset_time_width + tpset_time_offset ), and TPs are placed
+  // in TPSets based on the TP start time
+  //
+  // This loop assumes the input file is sorted by TP start time
+  for (std::string& fragment_path : fragment_paths) {
+    std::unique_ptr<dunedaq::daqdataformats::Fragment> frag = input_file->get_frag_ptr(fragment_path);
+    // Make sure this fragment is a TriggerPrimitive
+    if (frag->get_fragment_type() != dunedaq::daqdataformats::FragmentType::kTriggerPrimitive) continue;
```
ArturSztuc: Same as before, I would add a check to make sure the fragment with TP is specifically related with the Trigger, rather than e.g. readout-only.

ArturSztuc (approved):
LGTM. I would add an extra sanity check, however, as above in the review.

plugins/TriggerPrimitiveMaker.cpp
aeoranday: Legacy of old code. I can make that change before merging.

plugins/TriggerPrimitiveMaker.cpp
aeoranday: I did not refer to how TPSets are created in `readout`. The changes here were to make use of reading from HDF5 files on the legacy code. There were no logic or definition changes otherwise.

plugins/TriggerPrimitiveMaker.cpp
aeoranday: I can make that change before the merge.

Style question: should I be doing `if (...) continue;` (assuming short length) or (two lines)
```
if (...)
  continue;
```

ArturSztuc: That I care less about, I normally try to keep lines under ~70 characters and therefore mostly don't use `if(...) continue;`,  but the choice here is less intrusive :)

plugins/TriggerPrimitiveMaker.cpp
ArturSztuc: In that case this might not be the final PR for HDF5 reading. I think as much as possible we want to emulate how the readout creates TPSets, so that the replay app is really trying to replay identical conditions.

aeoranday: I agree. There will likely be more changes necessary to catch up to `readout`.

bieryAtFnal:
For formatting source code files, I suggest using `dbt-clang-format.sh <filename>`.

ArturSztuc (approved):
LGTM, thank you for all these changes!

PR #272, opened 2024-02-19
New replay app

MRiganSUSX:
New replay application that allows emulation of the trigger. 

Now reads in tpstream HDF5 files. 

It is based on the old replay app but re-coded as fddaqconf_gen changed a lot since. 

More details in README.


python/trigger/replay_tps/__main__.py
```diff
@@ -0,0 +1,542 @@
...
+    print("debug:", debug)
+    print("json_dir:", json_dir)
+    return
+
+def check_file_extension(input_file):
+    if input_file.lower().endswith(".hdf5"):
+        return
+    else: 
+        print("Invalid file provided. HDF5 tpstream file required.")
+        raise SystemExit(-1)
```
aeoranday: We may want to also include files that end with ".hdf5.copied" since many of the files we use get updated to this.

python/trigger/replay_tps/__main__.py
```diff
@@ -0,0 +1,542 @@
...
+
+    Args:
+        boot (_type_): _description_
+        readout (_type_): _description_
+        dataflow (_type_): _description_
+
+    Raises:
+        Exception: _description_
+        Exception: _description_
+        Exception: _description_
```
aeoranday, line 89: Very descriptive.

python/trigger/replay_tps/__main__.py
```diff
@@ -0,0 +1,542 @@
...
+    print("tpset-time-width:", tpset_time_width)
+    print("maximum-wait-time-us:", maximum_wait_time_us)
+    print("input-file:", input_file)
+    print("map-file:", map_file)
+    print("number-of-links:", number_of_links)
+    print("debug:", debug)
+    print("json_dir:", json_dir)
+    return
+
+def check_file_extension(input_file):
```
aeoranday, line 333: A check here that the HDF5 is a TPStream and not RawData would also be useful. Naive (and easily breakable) check could be on the file name, such as:
```python
if  '_raw_' in input_file:
    raise TypeError("Input file is a raw data file when it should be a tpstream file.")
```

A better solution would be to load the file and check that at least one fragment is a TimeSlice instead of a TriggerRecord. Ideally, files should not be mixing TimeSlices and TriggerRecords, so I think checking only one fragment would be sufficient. This does require `from hdf5libs import HDF5RawDataFile`.

python/trigger/replay_tps/__main__.py
```diff
@@ -0,0 +1,542 @@
+#!/usr/bin/env python3
+import click
+import math
+import os.path
```
aeoranday: The `os.path` import doesn't get used.

python/trigger/replay_tps/__main__.py
```diff
@@ -0,0 +1,542 @@
...
+        use_hsi_input=False,
+        DEBUG=debug)
+
+    return
+
+def dfo_apps(the_system, get_dfo_app, appconfig_df, daq_common, ru_descs, number_of_ru_streams, readout, dataflow, debug):
+    max_expected_tr_sequences = 1
+    for df_config in appconfig_df.values():
+        if df_config.max_trigger_record_window >= 1:
+            df_max_sequences = ((trigger.trigger_window_before_ticks + trigger.trigger_window_after_ticks) / df_config.max_trigger_record_window)
```
aeoranday, line 199: I'm confused how this line didn't produce errors: `trigger` isn't defined in this scope. I didn't have any issues while using the module, but maybe something to be aware of.

python/trigger/replay_tps/__main__.py
```diff
@@ -0,0 +1,542 @@
...
+    #--------------------------------------------------------------------------
+    # Check input files
+    #--------------------------------------------------------------------------
+    for each_file in input_file:
+        check_file_extension(each_file)
+
+    console.log("Loading dataflow config generator")
+    from daqconf.apps.dataflow_gen import get_dataflow_app
+    console.log("Loading readout config generator")
+    from fddaqconf.apps.readout_gen import FDReadoutAppGenerator
```
aeoranday: The `FDReadoutAppGenerator` import doesn't get used.

aeoranday (approved):
Use of the application is fine, but the data generated from `nanorc` afterwards is not. The number of TPs per TriggerRecord is significantly lower than per TimeSlice of the original file. Most TRs contain ~2 TPs with occasional jumps up to ~6 and infrequent jumps up to ~20. I'm suspicious that the issue is in TriggerPrimitiveMaker, but I have not looked into it.

Otherwise, I am able to read these TPs, TAs, and TCs and they do appear to match the configurations set otherwise; though, I only tested TAs and TCs with Prescale. All TPs were above the configured threshold.

While there are issues, I do believe that they are outside of this PR. Approved.

python/trigger/replay_tps/__main__.py
MRiganSUSX: included with next commit.

python/trigger/replay_tps/__main__.py
MRiganSUSX: I'm following the standards set by fddaqconf, to be in line ;) 
[fddaqconf_gen](https://github.com/DUNE-DAQ/fddaqconf/blob/production/v4/scripts/fddaqconf_gen#L93-L110)

python/trigger/replay_tps/__main__.py
MRiganSUSX: don't think we should rely on file names (barring file extensions, which have higher chance of being correct). names can change on purpose, by mistake, over time ...
agree loading and checking the file would be good, but since this is supposed to 'just' create the config, and leave the file handling to TPM, I think we should just leave it for now, and perhaps include with replay2.0 (or is this 2.0?) if we want

python/trigger/replay_tps/__main__.py
MRiganSUSX: removed

python/trigger/replay_tps/__main__.py
MRiganSUSX: I think the `trigger` object is just available in the global scope. But have passed it to this function in the next commit for completeness.

python/trigger/replay_tps/__main__.py
MRiganSUSX: removed

MRiganSUSX:
Thanks @aeoranday, I have addressed your comments.

python/trigger/replay_tps/__main__.py
aeoranday: I'm fine with leaving it for now. TPM does cite an access error when trying raw data and you've put to use TPStream in the documentation.

aeoranday (approved):
Changes look good to me. Further testing using non-Prescale algorithms produced results that were more in line with expectation. The previous claim I made on TP fragments having a low count was a misunderstanding on my part for how TP fragments are written, so running tests on non-Prescale helped me clear that and understand it better.

ArturSztuc (approved):
LGTM!

PR #274, opened 2024-02-22
Changed kTPCDefault to kSimpleThreshold in several test/unittest files

bieryAtFnal:
One of several additional changes needed to react to [the change from a single TP Algorithm name to several of them with more descriptive names](https://github.com/DUNE-DAQ/trgdataformats/pull/10/files).


aeoranday (approved):
LGTM.

Only `AlgorithmPlugins_test.cxx` failed the unit test, but this is for a known reason that an error is raised when trying to load an unavailable TA/TC (which was included as part of the test). I'll raise an issue for it so that it is at least documented.

PR #276, opened 2024-02-29
Updated AlgorithmPlugins Unit Test.

### LINKED ISSUE #275: AlgorithmPlugins Unit Test Raises Expected Error
The change in 16345ef to add errors for requesting a factory item that is not available was not propagated to the associated unit test. The test included a check that a `nullptr` is returned on such requests, so this now raises an error instead and fails the test.

https://github.com/DUNE-DAQ/trigger/blob/cf87d5e27c3c80d01fe0fb4d487e3ef54f922180/unittest/AlgorithmPlugins_test.cxx#L27

aeoranday:
This closes #275: `AlgorithmPlugins.hpp` was previously updated with a design to throw an error if a TA or TC factory attempted to build a maker that is not registered. The unit test still attempts to load a "fake" maker, but it now properly handles the error and uses that result as a success.

Testing was done by doing `dbt-build --unittest`.

### DIFF
--- a/unittest/AlgorithmPlugins_test.cxx
+++ b/unittest/AlgorithmPlugins_test.cxx
@@ -6,6 +6,7 @@
  */
 
 #include "trigger/AlgorithmPlugins.hpp"
+#include "trigger/Issues.hpp"
 
 /**
  * @brief Name of this test module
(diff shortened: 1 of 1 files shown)


ArturSztuc (approved):
LGTM

PR #279, opened 2024-03-06
Glm/testing

glehmannmiotto:
merge back intermediate set of fixes. still problems with TA fragments generation

### DIFF
--- a/src/TAProcessor.cpp
+++ b/src/TAProcessor.cpp
@@ -89,7 +89,7 @@ TAProcessor::conf(const appdal::ReadoutModule* conf)
     TLOG() << "Selected TC algorithm: " << algo->UID();
     std::shared_ptr<triggeralgs::TriggerCandidateMaker> maker = make_tc_maker(algo->class_name());
     nlohmann::json algo_json = algo->to_json(true);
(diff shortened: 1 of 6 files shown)


src/TAProcessor.cpp
```diff
@@ -87,11 +87,11 @@ TAProcessor::conf(const appdal::ReadoutModule* conf)
 
   for (auto algo : tc_algorithms)  {
     TLOG() << "Selected TC algorithm: " << algo->UID();
-    std::unique_ptr<triggeralgs::TriggerCandidateMaker> maker = make_tc_maker(algo->class_name());
+    std::shared_ptr<triggeralgs::TriggerCandidateMaker> maker = make_tc_maker(algo->class_name());
```
eflumerf, line 90: make_tc_maker still returns a unique_ptr, no?

glehmannmiotto: yes, but that is a typical pattern for a factory. It returns a unique pointer which is the more strict type of signature, but it is allowed to transform it right away to a shared ptr, if the user code requires it.

eflumerf (approved):

PR #280, opened 2024-03-07
Asztuc/tcmaking oks fix

ArturSztuc:
Adding new subscriber model that does `TriggerDatatype` -> `TDatatypeWrapper` conversion correctly, via 
`TDatatypeWrapper tdata(TriggerDatatype)` rather than with casts (the `TXWrapper(TriggerX _)` constructor must be called to fill the buffers correctly).

It feels a bit annoying to make whole new class with everything copy-pasted, maybe basing this class off `readoutlibs::DataSubscriberModel` and overloading `DataSubscriberModel::handle_payload` would be better here? That's why it's a draft PR for now, in case @glehmannmiotto has a better idea.

### DIFF
--- a/plugins/DataSubscriber.cpp
+++ b/plugins/DataSubscriber.cpp
@@ -13,6 +13,7 @@
 #include "readoutlibs/models/DataSubscriberModel.hpp"
 #include "trigger/HSISourceModel.hpp"
 #include "trigger/TPSetSourceModel.hpp"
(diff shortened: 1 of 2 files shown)


glehmannmiotto (approved):
Accepting it for now, so we can test extensively, but indeed we have to make better use of inheritance. So let's keep in mind that we need to revisit this before the release.

PR #284, opened 2024-03-19
Asztuc/max time over threshold

ArturSztuc:
Added TP filtering based on time over threshold at the `TPChannelFilter` level, as discussed in https://github.com/DUNE-DAQ/ehn1-operations-issues/issues/11. The default threshold for time-over-threshold filtering currently set to 10,000 ticks. Another PR to be added in daqconf.

Tests ran:
2. Used the trigger replay application to run with 2 configurations over tpstream from run `23582`:
   1. Filter out all the TPs with TOT lesser than 1,000 ticks
   2. Filter out all the TPs with TOT lesser than 50,000 ticks

First test successful, second test gave expected output. Showing below three graphs, each with a the distribution of time-over-threshold(s).

1. TOT for the full tpstream from run `23582`, before the changes. Maximum TOT is around 10,000, but most are below 2,000 (warning: red is log axis, gray is normal!)
tpstream_tot

2. TOT from all the TPs retrieved from `TPBuffers` after running through replay application with TOT filter of 1,000
tps_maxtot1000
 ticks. 

3. TOT from all the TPs retrieved from `TPBuffers` after running through replay application with TOT filter of 50,000
tps_maxtot50000

The reason why 3. doesn't cover entire range up to 10,000 as expected from 1., is because 1. is from a tp-stream file taken over a few hours, and I ran the replay application for 10 minutes.


plugins/TPChannelFilter.cpp
```diff
@@ -106,6 +106,12 @@ TPChannelFilter::channel_should_be_removed(int channel) const
   return false;
 }
 
+bool
+TPChannelFilter::remove_if_large_tot(int tot) const
+{
+  return (tot > m_conf.max_time_over_threshold) ? true : false;
+}
+
```
aeoranday: Why does this need the short-hand logic? Wouldn't `tot > m_conf.max_time_over_threshold` already give true or false?

schema/trigger/tpchannelfilter.jsonnet
```diff
@@ -6,6 +6,7 @@ local types = {
   bool: s.boolean("Boolean"),
   string : s.string("String", moo.re.ident,
     doc="A string field"),
+  ticks: s.number("ticks", dtype="i8"),
   
   conf : s.record("Conf", [
```
aeoranday: Very small thing. Would it be better to have `u8` instead of `i8`?

aeoranday (changes requested):
I've tested this and found all the TPs were compliant with the change. There are some small changes that I've commented on before approving.

plugins/TPChannelFilter.cpp
ArturSztuc: Agreed, I committed a change that simplifies this.

schema/trigger/tpchannelfilter.jsonnet
ArturSztuc: Agreed, I don't remember why I picked i8... changed it to u8 and it still works.

aeoranday (approved):
LGTM.

PR #285, opened 2024-03-20
Logging system rework

MRiganSUSX:
This was initiated by issues with trigger log files being too big (https://github.com/DUNE-DAQ/ehn1-operations-issues/issues/32)
There is accompanying PR in triggerals: https://github.com/DUNE-DAQ/triggeralgs/pull/59

This solution introduces new logging system with predefined levels. 

The idea is to:
- always see very important messages
- get important debug messages with low debug levels
- get increasingly more debug messages with higher debug levels, ordered by both presumed importance and also expected frequency
- the order of importance of objects is roughly TD > TC > TA > TP
- one line summary messages are kept as TLOG, so that they are always seen (ei MLT's message about # of TDs, TCs, Livetime)
- the levels are kept at consistent levels across modules, ie 'entering start method' is always the same level regardless of module (well, I tried) 

- I have added little 'tag' at the start of each TLOG(_DEBUG) to highlight the module
- changed some messages to warning/errors
- added few more details to some messages
- no (other) logic changes

Tested with configs from the instructions page & custom configs (RTCM, CTCM, ROI).
Also tested with integtests:
- minimal_system_quick_test.py
- fake_data_producer_test.py
- readout_type_scan.py
- tpstream_writing_test.py

One can change the debug level with:
```export DUNEDAQ_ERS_DEBUG_LEVEL=X```

### DIFF
--- a/include/trigger/Logging.hpp
+++ b/include/trigger/Logging.hpp
@@ -1,0 +1,34 @@
+/**
+ * @file Logging.hpp Common logging declarations in triggeralgs
+ *
+ * This is part of the DUNE DAQ , copyright 2020.
+ * Licensing/copyright details are in the COPYING file that you should have
+ * received with this code.
+ */
+#ifndef TRIGGER_INCLUDE_LOGGING_HPP_
+#define TRIGGER_INCLUDE_LOGGING_HPP_
+
+namespace dunedaq {
+namespace trigger {
+namespace logging {
+
+/**
+* @brief Common name used by TRACE TLOG calls from this package
+*/
+enum
+{
+  TLVL_VERY_IMPORTANT = 1,
+  TLVL_IMPORTANT      = 2,
+  TLVL_GENERAL        = 3,
+  TLVL_DEBUG_INFO     = 4,
+  TLVL_DEBUG_LOW      = 5,
+  TLVL_DEBUG_MEDIUM   = 10,
+  TLVL_DEBUG_HIGH     = 15,
+  TLVL_DEBUG_ALL      = 20
+};
+
+} // namespace logging
+} // namespace trigger
+} // namespace dunedaq
+
+#endif // TRIGGER_INCLUDE_LOGGING_HPP_
--- a/plugins/TABuffer.cpp
+++ b/plugins/TABuffer.cpp
@@ -127,7 +127,7 @@ TABuffer::do_work(std::atomic<bool>& running_flag)
     }
   } // while (running_flag.load())
 
-  TLOG(1) << "[TAB] " << get_name() << " exiting do_work() method. Received " << n_tas_received << " TAs " << " and " << n_requests_received << " data requests";
+  TLOG() << "[TAB] " << get_name() << " exiting do_work() method. Received " << n_tas_received << " TAs " << " and " << n_requests_received << " data requests";
 }
 
 } // namespace trigger
--- a/plugins/TABuffer.cpp
+++ b/plugins/TABuffer.cpp
@@ -7,11 +7,15 @@
  */
 
 #include "TABuffer.hpp"
+#include "trigger/Logging.hpp"
 
 #include "appfwk/DAQModuleHelper.hpp"
 #include "daqdataformats/SourceID.hpp"
 
 #include <string>
+
+using dunedaq::trigger::logging::TLVL_GENERAL;
+using dunedaq::trigger::logging::TLVL_DEBUG_ALL;
 
 namespace dunedaq {
 namespace trigger {
@@ -58,7 +62,7 @@ TABuffer::do_conf(const nlohmann::json& args)
   m_latency_buffer_impl->conf(args);
   m_request_handler_impl->conf(args);
 
-  TLOG_DEBUG(2) << get_name() + " configured.";
+  TLOG_DEBUG(TLVL_GENERAL) << "[TAB] " << get_name() + " configured.";
 }
 
 void
@@ -66,7 +70,7 @@ TABuffer::do_start(const nlohmann::json& args)
 {
   m_request_handler_impl->start(args);
   m_thread.start_working_thread("tabuffer");
-  TLOG_DEBUG(2) << get_name() + " successfully started.";
+  TLOG_DEBUG(TLVL_GENERAL) << "[TAB] "  << get_name() + " successfully started.";
 }
 
 void
@@ -75,7 +79,7 @@ TABuffer::do_stop(const nlohmann::json& args)
   m_thread.stop_working_thread();
   m_request_handler_impl->stop(args);
   m_latency_buffer_impl->flush();
-  TLOG_DEBUG(2) << get_name() + " successfully stopped.";
+  TLOG_DEBUG(TLVL_GENERAL) << "[TAB] " << get_name() + " successfully stopped.";
 }
 
 void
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

File: include/trigger/Logging.hpp [hunk 1]
```diff
@@ -0,0 +1,34 @@
...
+enum
+{
+  TLVL_VERY_IMPORTANT = 1,
+  TLVL_IMPORTANT      = 2,
+  TLVL_GENERAL        = 3,
+  TLVL_DEBUG_INFO     = 4,
+  TLVL_DEBUG_LOW      = 5,
+  TLVL_DEBUG_MEDIUM   = 10,
+  TLVL_DEBUG_HIGH     = 15,
+  TLVL_DEBUG_ALL      = 20
```
ArturSztuc, line 27: (copy-pasting from the other review)
`IMPORTANT` and `VERY_IMPORTANT` is strange naming (I guess `GENERAL` and `DEBUG_INFO` are too). Is the naming similar across the rest of the DAQ? The usual naming for different logging verbosity in the industry is:
* `TRACE` for everything, so `TLVL_DEBUG_ALL`
* `DEBUG` for the debugging messages (here it makes sense to have low, medium, high)
* `INFO` would be your `VERY_IMPORTANT`, `IMPORTANT` and/or `DEBUG_INFO`,
* `WARN` for warnings (we have different system for that),
* `ERROR` for errors (we have different system for that),
* `FATAL` for fatal errors that should crash (we have different system for that).

What are the namings in the rest of the DAQ?

plugins/TABuffer.cpp
```diff
@@ -100,30 +104,30 @@ TABuffer::do_work(std::atomic<bool>& running_flag)
...
+      //TLOG_DEBUG(TLVL_DEBUG_ALL) << "[TAB] Handled data request, occupancy is: " << m_latency_buffer_impl->occupancy();
     }
 
     if (!popped_anything) {
       std::this_thread::sleep_for(std::chrono::milliseconds(10));
     }
   } // while (running_flag.load())
 
-  TLOG() << get_name() << " exiting do_work() method. Received " << n_tas_received << " TAs " << " and " << n_requests_received << " data requests";
+  TLOG(1) << "[TAB] " << get_name() << " exiting do_work() method. Received " << n_tas_received << " TAs " << " and " << n_requests_received << " data requests";
```
ArturSztuc: Isn't `TLOG(1)` similar/the same as `TLOG_DEBUG(TLVL_VERY_IMPORTANT)`? Or should it just ve `TLOG(TLVL_VERY_IMPORTANT)`?

plugins/TPSetBufferCreator.cpp
```diff
@@ -348,7 +355,7 @@ TPSetBufferCreator::do_work(std::atomic<bool>& running_flag)
     }
   } // end while(running_flag.load())
 
-  TLOG() << get_name() << ": Exiting the do_work() method: received " << addedCount << " Sets and " << requestedCount
+  TLOG(1) << "[TPSetBufferCreator] " << get_name() << ": Exiting the do_work() method: received " << addedCount << " Sets and " << requestedCount
```
ArturSztuc: Ditto re. `TLOG(1)` vs `TLVL_VERY_IMPORTANT`

ArturSztuc (changes requested):
Similarly to https://github.com/DUNE-DAQ/triggeralgs/pull/59, I find the logging names strange (and somewhat confusing, e.g. `TLVL_GENERAL` can be easily interpreted as == 0 or 1.

One thing I forgot to comment on https://github.com/DUNE-DAQ/triggeralgs/pull/59, why all the logging messages now have e.g. [TGM], [MLT] etc? In the actual logging message the name of the class is appended to the message already, so this seems redundant and makes the message longer... e.g. old message:

`INFO [void dunedaq::trigger::ModuleLevelTrigger::do_start(const nlohmann::json&) at /nfs/home/asztuc/workspace/releases/rel_NFD_PROD4_240320_A9/sourcecode/trigger    /plugins/ModuleLevelTrigger.cpp:187] Start of run 1111`

Will now be changed to:
`INFO [void dunedaq::trigger::ModuleLevelTrigger::do_start(const nlohmann::json&) at /nfs/home/asztuc/workspace/releases/rel_NFD_PROD4_240320_A9/sourcecode/trigger    /plugins/ModuleLevelTrigger.cpp:187] [MLT] Start of run 1111`
So we now get the extra `[MLT]` but we already know it's `ModuleLevelTrigger` anyway?

bieryAtFnal:
I have similar questions as Artur.

I will also add "what is intended with TLOG(1)?"  That is different than TLOG_DEBUG(1), and I'd like to understand what the goal of TLOG(1) is, compare to what it actually does.

Also, I'd like to understand why this large change set is targeted to the v4 development line.  That seems orthogonal to the spirit of making the minimal changes that we need for NP04 running.  And, my sense is that it will make future merges to the v5 development line harder.

I'd appreciate some discussion of these topics on today's Release Planning call.

ArturSztuc:
@bieryAtFnal 
> Also, I'd like to understand why this large change set is targeted to the v4 development line. That seems orthogonal to the spirit of making the minimal changes that we need for NP04 running. And, my sense is that it will make future merges to the v5 development line harder.
> 
> I'd appreciate some discussion of these topics on today's Release Planning call.

I think it somewhat makes sense to have at least part of it in v4 because of the large logging output from the trigger. If we're running v4 on PD2, we will likely make far more TAs than TCs, and as result get very large logging output from the trigger (potentially 100-1000 lines a second).
I guess this work could be split into two: small change to v4 (one-two lines) to reduce the logging from TA makers, and full logging change for v5, but that could indeed be discussed at the release planning call.

include/trigger/Logging.hpp
MRiganSUSX: (copy-paste response from triggeralgs)
This naming is completely custom. As far as I know, there is no consistency across dunedaq. There are some repos that have a version of this, example [in-file enum here](https://github.com/DUNE-DAQ/flxlibs/blob/production/v4/src/CardWrapper.cpp#L26-L31) and [separare logging header here](https://github.com/DUNE-DAQ/readoutlibs/blob/production/v4/include/readoutlibs/ReadoutLogging.hpp#L18-L29). Other places simply use integers [example here](https://github.com/DUNE-DAQ/hsilibs/blob/production/v4/plugins/HSIReadout.cpp#L250), and these do go to high values ([example here](https://github.com/DUNE-DAQ/iomanager/blob/production/v4/src/network/ConfigClient.cpp#L89)).

I just made that list according to what made sense to me, and wanted it to (a) be granural at the lower levels and (b) roughly cover the available space for levels. Open to changes. As you mentioned, we are not using standards from the industry (or at least not closely).

plugins/TABuffer.cpp
MRiganSUSX: I actually don't think `TLOG` uses any levels, it will just always print. There are old instances across trigger and triggeralgs where the `TLOG` is given a level but I think this is a mistake/oversight from times past. I will remove these numbers (I used 1 across to make it consistent, but a consistent misusage is still misusage).

plugins/TPSetBufferCreator.cpp
MRiganSUSX: will remove.

MRiganSUSX:
Thanks both for the comments. Response here:

- `why all the logging messages now have e.g. [TGM], [MLT] etc?`
This is just something I've seen Alessandro use before: [example here](https://github.com/DUNE-DAQ/triggeralgs/blob/production/v4/src/TriggerActivityMakerHorizontalMuon.cpp#L58) and [here](https://github.com/DUNE-DAQ/triggeralgs/blob/production/v4/src/TriggerCandidateMakerHorizontalMuon.cpp#L70) and I liked it. I think especially for debugging messages it makes sense as you can simply filter by these tags when looking at logs. It makes it easy to follow how objects pass through the flow without the need to look at the convoluted way the tlog msg is introduced (this: `void dunedaq::trigger::ModuleLevelTrigger::do_start(const nlohmann::json&) at /nfs/home/asztuc/workspace/releases/rel_NFD_PROD4_240320_A9/sourcecode/trigger    /plugins/ModuleLevelTrigger.cpp`), now we have [MLT]. You are correct it is not adding new information, we already have access to the filename, it's just easy to read/parse.

- `what is intended with TLOG(1)?`
As mentioned in a response to Artur's comment above, this is my mistake trying to follow what was already present across trigger(algs). The only thing I wanted to achieve now is to have a consistent number across the messages in the trigger repos. I will change all of these to TLOG(). 

- `Also, I'd like to understand why this large change set is targeted to the v4 development line.`
You are right that the issue that started this could have been corrected with few lines of code. However, I think a rework is appropriate at this point. There was no consistency across trigger before. These changes do not affect the functionality/logic of the repos, and (hopefully) add more information/consistency and make things easier to read/follow. I think this is helpful for Protodunes as well. 

While I'm not strongly pushing for any of this, I think it is helpful without high probability of harming functionality.

MRiganSUSX:
Fixed the TLOG(x) with commit above.

ArturSztuc (approved):

PR #287, opened 2024-03-22
Asztuc/trigger neutron source

ArturSztuc:
Changes to allow for pulsed neutron source calibration input and conversion to it's individual TriggerCandidate type. New TriggerCandidate type named `kNeutronSourceCalib = 10`.

Dependent on PRs in
- daqconf:  https://github.com/DUNE-DAQ/daqconf/pull/437
- trgdataformats: https://github.com/DUNE-DAQ/trgdataformats/pull/13

Ended up with a small overhaul of `TimingTriggerCandidateMaker` configuration, so it can accept any number of HSIEvent signal types. Moreso, these signal types can now be mapped to different TriggerCandidate Types at config time (by providing their string names), by supplying signal map in `daqconf.json` , e.g.:
```
"trigger":
   "ttcm_input_map": [
       {
          "signal": 0,
          "tc_type_name": "kTiming",
          "time_before": 1000,
          "time_before": 2000
       },
       {
          "signal": 1,
          "tc_type_name": "kTiming",
          "time_before": 1000,
          "time_before": 2000
       },
       {
          "signal": 3,
          "tc_type_name": "kNeutronSourceCalib",
          "time_before": 1000,
          "time_before": 2000
       }
```

By default config will create signals `[0, 1, 2, 3]`, all with `kTiming` as TC-type.

Tests ran:
2. Default config generation, checking if signal map for `[0, 1, 2, 3]` exists and tuned to `kTiming` in the config, ran offline with fake hsi on, and seen `kTiming` TCs in the output hdf5.
3. Ran offline with custom configuration to tune one of the signals to the new `kNeutronSourceCalib` TriggerCandidate type.  Seen output like:
   ``` 
        Trigger_Candidate fragment with SourceID Trigger_0x00000003 from subdetector DAQ has size = 128
                Readout window before = 2000, after = 2000
                TC type =  (10), TC algorithm = 2, number of TAs = 0
                Start time = 79554165758489587, end time = 79554165758493587, and candidate time = 79554165758491587
   ``` 
   where TC type 10 corresponds to the new `kNeutronSourceCalib`.

### DIFF
--- a/include/trigger/Issues.hpp
+++ b/include/trigger/Issues.hpp
@@ -189,6 +189,14 @@ ERS_DECLARE_ISSUE_BASE(trigger,
                        "Factory could not find requested item " << item << ".",
                        ((std::string)item),
                        ERS_EMPTY)
+
+ERS_DECLARE_ISSUE_BASE(trigger,
+                       ConfigurationProblem,
+                       appfwk::GeneralDAQModuleIssue,
+                       "Configuration error: " << item,
+                       ((std::string)item),
+                       ERS_EMPTY)
+
 } // namespace dunedaq
 
 #endif // TRIGGER_INCLUDE_TRIGGER_ISSUES_HPP_
--- a/include/trigger/Issues.hpp
+++ b/include/trigger/Issues.hpp
@@ -204,6 +204,7 @@ ERS_DECLARE_ISSUE_BASE(trigger,
                        TTCMConfigurationProblem,
                        appfwk::GeneralDAQModuleIssue,
                        "Configuration error: " << item,
+                       ((std::string)name),
                        ((std::string)item),
                        ERS_EMPTY)
 
--- a/include/trigger/Issues.hpp
+++ b/include/trigger/Issues.hpp
@@ -201,7 +201,7 @@ ERS_DECLARE_ISSUE_BASE(trigger,
                        ERS_EMPTY)
 
 ERS_DECLARE_ISSUE_BASE(trigger,
-                       ConfigurationProblem,
+                       TTCMConfigurationProblem,
                        appfwk::GeneralDAQModuleIssue,
                        "Configuration error: " << item,
                        ((std::string)item),
--- a/plugins/TimingTriggerCandidateMaker.cpp
+++ b/plugins/TimingTriggerCandidateMaker.cpp
@@ -75,7 +75,7 @@ TimingTriggerCandidateMaker::do_conf(const nlohmann::json& config)
   for (auto hsi_input : params.hsi_configs) {
     triggeralgs::TriggerCandidate::Type type;
     type = static_cast<triggeralgs::TriggerCandidate::Type>(
-        dunedaq::trgdataformats::string_to_fragment_type_value(hsi_input.tc_trigger_type));
+        dunedaq::trgdataformats::string_to_fragment_type_value(hsi_input.tc_type_name));
     if (type == triggeralgs::TriggerCandidate::Type::kUnknown) {
       throw ConfigurationProblem(ERS_HERE,
           "Unknown TriggerCandidate supplied to TTCM HSI map");
@@ -90,7 +90,7 @@ TimingTriggerCandidateMaker::do_conf(const nlohmann::json& config)
                                               hsi_input.time_before,
                                               hsi_input.time_after };
 
-    TLOG() << "TTCM will convert HSI signal id: " << hsi_input.signal << " to TC type: " << hsi_input.tc_trigger_type;
+    TLOG() << "TTCM will convert HSI signal id: " << hsi_input.signal << " to TC type: " << hsi_input.tc_type_name;
   }
 
   if (m_detid_offsets_map.empty()) {
--- a/plugins/TimingTriggerCandidateMaker.cpp
+++ b/plugins/TimingTriggerCandidateMaker.cpp
@@ -72,7 +72,7 @@ TimingTriggerCandidateMaker::HSIEventToTriggerCandidate(const dfmessages::HSIEve
     candidate.time_end   = data.timestamp + m_hsisignal_map[data.signal_map].time_after; // time_end,
     candidate.time_candidate = data.timestamp;
     // throw away bits 31-16 of header, that's OK for now
-    candidate.detid = { static_cast<triggeralgs::detid_t>(signal) }; // NOLINT(build/unsigned)
+    candidate.detid = data.header; // NOLINT(build/unsigned)
 
     candidate.type = m_hsisignal_map[signal].type; // type,
 
@@ -97,12 +97,12 @@ TimingTriggerCandidateMaker::do_conf(const nlohmann::json& config)
     type = static_cast<triggeralgs::TriggerCandidate::Type>(
         dunedaq::trgdataformats::string_to_fragment_type_value(hsi_input.tc_type_name));
     if (type == triggeralgs::TriggerCandidate::Type::kUnknown) {
-      throw TTCMConfigurationProblem(ERS_HERE,
+      throw TTCMConfigurationProblem(ERS_HERE, get_name(),
           "Unknown TriggerCandidate supplied to TTCM HSI map");
     }
 
     if (m_hsisignal_map.count(hsi_input.signal)) {
-      throw TTCMConfigurationProblem(ERS_HERE,
+      throw TTCMConfigurationProblem(ERS_HERE, get_name(),
           "Supplied more than one of the same hsi signal ID to TTCM HSI map");
     }
 
@@ -114,7 +114,7 @@ TimingTriggerCandidateMaker::do_conf(const nlohmann::json& config)
   }
 
   if (m_hsisignal_map.empty()) {
-      throw TTCMConfigurationProblem(ERS_HERE,
+      throw TTCMConfigurationProblem(ERS_HERE, get_name(),
           "Created TTCM, but supplied an empty signal map!");
   }
 
--- a/plugins/TimingTriggerCandidateMaker.cpp
+++ b/plugins/TimingTriggerCandidateMaker.cpp
@@ -45,8 +45,8 @@ TimingTriggerCandidateMaker::HSIEventToTriggerCandidate(const dfmessages::HSIEve
   } else {
     if (m_detid_offsets_map.count(data.signal_map)) {
       // clang-format off
-      candidate.time_start = data.timestamp - m_detid_offsets_map[data.signal_map].first;  // time_start
-      candidate.time_end   = data.timestamp + m_detid_offsets_map[data.signal_map].second; // time_end,
+      candidate.time_start = data.timestamp - m_detid_offsets_map[data.signal_map].time_before;  // time_start
+      candidate.time_end   = data.timestamp + m_detid_offsets_map[data.signal_map].time_after; // time_end,
       // clang-format on    
     } else {
       throw dunedaq::trigger::SignalTypeError(ERS_HERE, get_name(), data.signal_map);
@@ -56,12 +56,7 @@ TimingTriggerCandidateMaker::HSIEventToTriggerCandidate(const dfmessages::HSIEve
   // throw away bits 31-16 of header, that's OK for now
   candidate.detid = { static_cast<triggeralgs::detid_t>(data.signal_map) }; // NOLINT(build/unsigned)
 
-  if (data.signal_map == m_neutron_source_signal) {
-    candidate.type = triggeralgs::TriggerCandidate::Type::kNeutronSourceCalib;
-  }
-  else {
-    candidate.type = triggeralgs::TriggerCandidate::Type::kTiming;
-  }
+  candidate.type = m_detid_offsets_map[data.signal_map].type; // type,
 
   candidate.algorithm = triggeralgs::TriggerCandidate::Algorithm::kHSIEventToTriggerCandidate;
   candidate.inputs = {};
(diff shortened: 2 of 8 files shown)


strilov:
looks pretty good to me; couple of comments/questions:

* perhaps the variable `m_detid_offsets_map` should be called something else now?
* the signal map field in HSIEvent is still treated as an integer; will the bit-map treatment be added in a different PR?
* for the defaults, I'd just have bit 0 be kTiming (or better kFakeHSI (if that's added as a TC type))
* With these changes, I think the pass-through mode no longer needed?

ArturSztuc:
@strilov :
* Agreed, changed to `m_hsisignal_map`.
* I was going to do it in a separate PR, but corrected it here now. Re-tested, and I'm still getting the same triggers etc. as before, but now the maker first finds which bits are triggered, and creates a TC for each bit.
* The default configuration sets 0 to `kTiming`, so that's already done. We should discuss what other types we  might want, I assume kFakeHSI is not the only one we might want... that could be added in a separate PR with the rest of trigger types that we might want to include for DTS/FakeHSI etc.
* I think it was no longer needed a while back, @MRiganSUSX says it was originally a hack for iceberg that's no longer needed. I removed it in the last commit

MRiganSUSX:
Regarding the `hsi_passthrough`, I know it was used as a bypass for Iceberg set-up, but I'm not sure whether it is still needed or not.

include/trigger/Issues.hpp
```diff
@@ -189,6 +189,14 @@ ERS_DECLARE_ISSUE_BASE(trigger,
                        "Factory could not find requested item " << item << ".",
                        ((std::string)item),
                        ERS_EMPTY)
+
+ERS_DECLARE_ISSUE_BASE(trigger,
+                       ConfigurationProblem,
```
MRiganSUSX: Unless this will be used globally it should be renamed to more easily point to the issue. IE `UnalignedHeartbeat` is very precise/clear, etc...

include/trigger/Issues.hpp
```diff
@@ -189,6 +189,14 @@ ERS_DECLARE_ISSUE_BASE(trigger,
...
+                       appfwk::GeneralDAQModuleIssue,
+                       "Configuration error: " << item,
+                       ((std::string)item),
```
MRiganSUSX, line 208: seems like this is missing the `((std::string)name),` part which passes name of the module to the errors platform (see other issues)

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -37,48 +38,87 @@ TimingTriggerCandidateMaker::TimingTriggerCandidateMaker(const std::string& name
...
 
-triggeralgs::TriggerCandidate
+template<typename T> 
+std::vector<int> 
+TimingTriggerCandidateMaker::GetTriggeredBits(T signal_map)
+{
+  std::vector<int> triggered_bits;
+  int nbits = std::numeric_limits<T>::digits;
+  for (int i = 0; i < nbits; ++i) {
+    if ((signal_map >> i) & 1 ) {
```
MRiganSUSX, line 48: I had to do a similar (same?) thing for the CTB plugin, using bitset instead of vector and looping over (see: https://github.com/DUNE-DAQ/trigger/blob/mrigan/ctb_integration/plugins/CTBTriggerCandidateMaker.cpp#L49C3-L49C40). Just another implementation possibility if you'd like.

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -37,48 +38,87 @@ TimingTriggerCandidateMaker::TimingTriggerCandidateMaker(const std::string& name
...
-  candidate.type = triggeralgs::TriggerCandidate::Type::kTiming;
 
-  candidate.algorithm = triggeralgs::TriggerCandidate::Algorithm::kHSIEventToTriggerCandidate;
-  candidate.inputs = {};
+    triggeralgs::TriggerCandidate candidate;
+    candidate.time_start = data.timestamp - m_hsisignal_map[data.signal_map].time_before;  // time_start
+    candidate.time_end   = data.timestamp + m_hsisignal_map[data.signal_map].time_after; // time_end,
+    candidate.time_candidate = data.timestamp;
+    // throw away bits 31-16 of header, that's OK for now
+    candidate.detid = { static_cast<triggeralgs::detid_t>(signal) }; // NOLINT(build/unsigned)
```
MRiganSUSX: Is this still needed (storing data to `detid` that is not detid) ?

schema/trigger/timingtriggercandidatemaker.jsonnet
```diff
@@ -3,52 +3,27 @@ local ns = "dunedaq.trigger.timingtriggercandidatemaker";
...
-			doc="Example 2"),
-		s.field("hsi_trigger_type_passthrough", self.hsi_tt_pt, doc="Option to override the trigger type values"),
-                s.field("prescale", self.count_t, default=1, doc="Option to prescale TTCM TCs")
+  count_t : s.number("count_t", "u8", nc(minimum=1), doc="Counter"),
+
+  hsi_input: s.record("hsi_input", [
+    s.field("signal",   self.signal_type_t,   default=0),
+    s.field("tc_type_name", self.trigger_type_t,  default="kTiming"),
+    s.field("time_before",  self.time_t,          default=10000, doc="Readout time before time stamp"),
+    s.field("time_after",   self.time_t,          default=20000, doc="Readout time after time stamp"),
```
MRiganSUSX, line 18: again nitpicking: any reason these defaults differ from the ones in daqconf ?

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -37,48 +38,87 @@ TimingTriggerCandidateMaker::TimingTriggerCandidateMaker(const std::string& name
...
-  m_hsi_pt_after = params.s0.time_after;
+
+  // Fill the internal hsi signal map. The map contains signal ID, TC type to
+  // create, and the readout time window.
+  for (auto hsi_input : params.hsi_configs) {
+    triggeralgs::TriggerCandidate::Type type;
+    type = static_cast<triggeralgs::TriggerCandidate::Type>(
+        dunedaq::trgdataformats::string_to_fragment_type_value(hsi_input.tc_type_name));
+    if (type == triggeralgs::TriggerCandidate::Type::kUnknown) {
+      throw ConfigurationProblem(ERS_HERE,
```
MRiganSUSX: ditto regarding the issue name

plugins/TimingTriggerCandidateMaker.cpp
```diff
@@ -37,48 +38,87 @@ TimingTriggerCandidateMaker::TimingTriggerCandidateMaker(const std::string& name
...
+    if (m_hsisignal_map.count(hsi_input.signal)) {
+      throw ConfigurationProblem(ERS_HERE,
+          "Supplied more than one of the same hsi signal ID to TTCM HSI map");
+    }
+
+    m_hsisignal_map[hsi_input.signal] = { type,
+                                          hsi_input.time_before,
+                                          hsi_input.time_after };
+
+    TLOG() << "[TTCM] will convert HSI signal id: " << hsi_input.signal << " to TC type: " << hsi_input.tc_type_name;
```
MRiganSUSX, line 113: do we want this to be a TLOG instead of some low level debug ?

schema/trigger/timingtriggercandidatemaker.jsonnet
```diff
@@ -3,52 +3,27 @@ local ns = "dunedaq.trigger.timingtriggercandidatemaker";
...
+    ]),
+
+  hsi_inputs: s.sequence("hsi_inputs", self.hsi_input),
+
+  conf: s.record("Conf", [
+    s.field("hsi_configs", self.hsi_inputs, [], doc="List of the input HSI configurations"),
+		s.field("hsi_trigger_type_passthrough", self.hsi_tt_pt, default=false, doc="Option to override the trigger type values"),
```
MRiganSUSX: if we are not using the passthrough it should be removed from the config

MRiganSUSX (commented):
Logic looks correct. I've left few comments.

include/trigger/Issues.hpp
ArturSztuc: The idea is that it could be used more globally, as a general "there was an error" and "here's a message that dev wanted to include here". I can make it more precise to?

plugins/TimingTriggerCandidateMaker.cpp
ArturSztuc: Probably not, I just didn't want to make this change here. I can change to header instead?

schema/trigger/timingtriggercandidatemaker.jsonnet
ArturSztuc: I copy-pasted these from the old config. In the old config these were set as above, but the `trigger_gen` / `get_trigger_app` overwrites these with the first entry anyway. I'm not sure why, it was done that way, but I was trying to preserve it in case it was on purpose. Do you think it should be changed?

plugins/TimingTriggerCandidateMaker.cpp
ArturSztuc: I just wanted a very generic but costomizable "there's a config issue" message here, should I make a separate issue for this?

plugins/TimingTriggerCandidateMaker.cpp
ArturSztuc: I wanted all the "config" and "do stop" "do start" etc. to be simple TLOG(), because they will not be executed often. Can discuss further/change if needed though?

schema/trigger/timingtriggercandidatemaker.jsonnet
ArturSztuc: Done, changed in the `daqconf` too

plugins/TimingTriggerCandidateMaker.cpp
ArturSztuc: Initially I had something pretty much identical, but decided to abstract out to not have the hard coded 32 and keep the rest semi-clean

include/trigger/Issues.hpp
MRiganSUSX: I think the idea of the issue system was to be very specific (certainly the case for currently existing issues). Maybe if you want it to be used globally (for that plugin) add HSI/timing or some reference to the names?

plugins/TimingTriggerCandidateMaker.cpp
MRiganSUSX: It seems slower to me from a glance, but I have not tested/compared the implementations. I'm happy with both though.

plugins/TimingTriggerCandidateMaker.cpp
MRiganSUSX: I think the idea was to drop passing this information through. We are using the header in other places - although I'm not sure that is used/useful.

schema/trigger/timingtriggercandidatemaker.jsonnet
MRiganSUSX: I think the original purpose of s0, s1 ... was to have different readout windows, but maybe @strilov could confirm.

plugins/TimingTriggerCandidateMaker.cpp
MRiganSUSX: I'm not sure whether there are rules for how the issues are to be defined. I *think* from looking at existing ones that the idea is to be as specific as possible - including passing information about the variables involved. I'd be happy with a global config issue, as long as it is contained for the plugin and has a reference to it.

plugins/TimingTriggerCandidateMaker.cpp
MRiganSUSX: The TLOG rework has set these 'one time messages' as GENERAL debug level, I think. Don't think it matters too much for these. I was mostly asking about the 'will convert HSI signal' message as that one is expected a lot more often.

include/trigger/Issues.hpp
ArturSztuc: Changed, but only to move this forward -- I still don't like it.
I did have a look at other Issue.hpp, e.g. in appfwk and readout, and many of the issues are very generic there so they can be reused across the whole repository, rather than only in one class.
Adding things like names of classes is redundant -- this is going to be added automatically anyway by the logger.

plugins/TimingTriggerCandidateMaker.cpp
ArturSztuc: Changed, but as before, I don't like that change.

plugins/TimingTriggerCandidateMaker.cpp
ArturSztuc: This message is only in the configuration, not in the conversion itself. 
So when receiving HSI signal, completely different lines of code will be triggered. 
This one will only be triggered once, when running `conf` through nanorc.

include/trigger/Issues.hpp
ArturSztuc: Done.

plugins/TimingTriggerCandidateMaker.cpp
ArturSztuc: Changed to `header`, which we are not using anywhere else, but maybe will be useful for others to distinguish different "kTiming" signals.

include/trigger/Issues.hpp
MRiganSUSX: I think having a global 'IssueAtConfig' is better (with you on this), I'm just pointing out that this is not the case for the Issues we have currently available, and therefore - for consistency - we should apply the same method now until a full rework of issues is done (probably not best time to do now).

MRiganSUSX (approved):
LGTM.
I think there are some things we should not forget about (default values for s0, s1; rework of issues, etc) but not urgent right now.

PR #288, opened 2024-03-25
CTB: Trigger integration

MRiganSUSX:
Implementation of the CTB signals into trigger.

There is a new plugin `CTBTriggerCandidateMaker` based on the `TimingTriggerCandidateMaker`. The plugin is 'standalone' (not part of the TP flow) and connects to the MLT (and to TC Buffer). 
This plugin receives HSIEvents and converts them to TCs. New TC types have been added to reflect this (pybinds included).
The creation of the TCs works in a loop as it is possible to receive a signal with multiple bits on, each bit corresponding to a (to-be-made) TC of a different type. 
There is a map of signal bits to TC types in the header file.
There is an error thrown if the received bit is not known (not in the map).
There is functional opmon for the hsi events and TC objects. 

Because there are now many new TC types, several other wrappers had to be extended. This includes the `bitsets` in the MLT as well as the `trigger_type` in the TriggerDecision message. This is also propagated all the way to the TriggerRecordHeaderData. As a consequence the TRHD is now bigger (!) and padding had to be adjusted to accommodate for this.

!!! IMPORTANTLY, the version of both TriggerCandidateData and TriggerRecordHeaderData were increased to reflect these big changes. !!!
-

The CTB plugin has 3 configurable parameters:
	"ctb_prescale": 1,                 <- we can prescale the events
	"ctb_time_before": 1000,     <- extension to the readout window before hsievent timestamp
	"ctb_time_after": 1000,        <- extension to the readout window after hsievent timestamp

This passes the integration tests: minimal_system_quick_test.py, readout_type_scan.py, fake_data_producer_test.py.

From custom test, I was able to use fake_hsi to get HSI events of different signals.
These are picked by the CTB:
```
[CTB] Converting HSI event, signal: 8880680
```
Bits are extracted:
```
BITS: 00000000100001111000001000101000
```
and appropriate TCs are created:
```
this bit: 3
TC type: 13
```
These are received at MLT:
```
 Got TC of type 13, timestamp 79554162131545930, start/end 79554162131544930/79554162131546930
```
As was the case before, the trigger_type inside the HDF5 file corresponds to the trigger_type sent by the MLT.

### DIFF
--- a/plugins/CTBTriggerCandidateMaker.cpp
+++ b/plugins/CTBTriggerCandidateMaker.cpp
@@ -1,0 +1,200 @@
+/**
+ * @file CTBTriggerCandidateMaker.cpp
+ *
+ * This is part of the DUNE DAQ Application Framework, copyright 2020.
+ * Licensing/copyright details are in the COPYING file that you should have
+ * received with this code.
+ */
+
+#include "CTBTriggerCandidateMaker.hpp"
+
+#include "appfwk/DAQModuleHelper.hpp"
+#include "trgdataformats/Types.hpp"
+#include "trigger/TriggerCandidate_serialization.hpp"
+#include "iomanager/IOManager.hpp"
+#include "rcif/cmd/Nljs.hpp"
+
+#include <regex>
+#include <string>
+
+namespace dunedaq {
+namespace trigger {
+
+CTBTriggerCandidateMaker::CTBTriggerCandidateMaker(const std::string& name)
+  : DAQModule(name)
+  , m_output_queue(nullptr)
+  , m_queue_timeout(100)
+{
+
+  register_command("conf", &CTBTriggerCandidateMaker::do_conf);
+  register_command("start", &CTBTriggerCandidateMaker::do_start);
+  register_command("stop", &CTBTriggerCandidateMaker::do_stop);
+  register_command("scrap", &CTBTriggerCandidateMaker::do_scrap);
+}
+
+triggeralgs::TriggerCandidate
+CTBTriggerCandidateMaker::HSIEventToTriggerCandidate(const dfmessages::HSIEvent& data)
+{
+  triggeralgs::TriggerCandidate candidate;
+  if (m_hsi_passthrough == true) {
+    TLOG_DEBUG(3) << "HSI passthrough applied, modified readout window is set";
+    candidate.time_start = data.timestamp - m_hsi_pt_before;
+    candidate.time_end = data.timestamp + m_hsi_pt_after;
+  } else {
+    if (m_detid_offsets_map.count(data.signal_map)) {
+      // clang-format off
+      candidate.time_start = data.timestamp - m_detid_offsets_map[data.signal_map].first;  // time_start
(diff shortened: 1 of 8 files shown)


MRiganSUSX:
Accompanying PRs:
https://github.com/DUNE-DAQ/trgdataformats/pull/16
https://github.com/DUNE-DAQ/daqdataformats/pull/53
https://github.com/DUNE-DAQ/daqconf/pull/441
https://github.com/DUNE-DAQ/fddaqconf/pull/30

bieryAtFnal (approved):
My testing so far looks good.

One note is that we will need to change hdf5libs/scripts/hdf5_dump.py to take into account the larger size for TriggerRecordHeader::trigger_type, but I want to ask people about versioning inside that script, and I don't want to hold up the bulk of the changes in the meantime.

plugins/CTBTriggerCandidateMaker.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+      if (m_HLT_TC_map.count(i)) {
+        TLOG_DEBUG(TLVL_DEBUG_ALL) << "[CTB] TC type: " << static_cast<int>(m_HLT_TC_map[i]);
+
+        triggeralgs::TriggerCandidate candidate;
+        candidate.time_candidate = data.timestamp;
+        candidate.time_start = data.timestamp - m_time_before;
+        candidate.time_end = data.timestamp + m_time_after;
+        // candidate.detid = 1;
+        candidate.detid = data.header;
```
ArturSztuc, line 65: I think this is probably the best way of dealing with "detid" here... `HSIEvent::header` tells us which device the event came from

plugins/CTBTriggerCandidateMaker.cpp
```diff
@@ -0,0 +1,204 @@
...
+CTBTriggerCandidateMaker::HSIEventToTriggerCandidate(const dfmessages::HSIEvent& data)
+{
+  TLOG_DEBUG(TLVL_DEBUG_MEDIUM) << "[CTB] Converting HSI event, signal: " << data.signal_map;
+
+  std::vector<triggeralgs::TriggerCandidate> candidates;
+  std::bitset<32> bits(data.signal_map);
+  TLOG_DEBUG(TLVL_DEBUG_HIGH) << "[CTB] BITS: " << bits;
+
+  for (size_t i = 0; i < bits.size(); ++i) {
+    if (bits.test(i)) {
```
ArturSztuc, line 53: Using guard clauses instead, e.g. 
```
if (!bits.test(i)) {
    continue;
}
```

would mean not having large nested `if`s. More here: https://refactoring.guru/replace-nested-conditional-with-guard-clauses

plugins/ModuleLevelTrigger.hpp
ArturSztuc, line 1: Wonder if the number of bits could be saved somewhere and used throughout, but I guess we won't be changing it too often...

ArturSztuc (approved):
LGTM. I have some comments, but nothing to hold up the merge.

ArturSztuc:
Testing the CTB logic to be discussed at today's v4 meeting

PR #289, opened 2024-03-27
Fixed compiler error in Issues.hpp and cleaned up a couple of typos t…

bieryAtFnal:
…hat would have caused odd messages to be printed out

I removed the extra ERS_EMPTY parameter in the TTCMConfigurationParameter Issue declaration.  This was causing a compiler error.

I also added the "size" parameter to the message string in the TCTimestampsSizeError Issue.  That seemed appropriate  (why have it if it isn't displayed?)

and, I changed the declaration type for MissingFactoryItemError since it is not being passed a get_name() value in the code that makes use of it.


ArturSztuc (approved):
LGTM

PR #290, opened 2024-03-27
Replay fix to (not) use CTB & fake HSI

ArturSztuc:
Replay application became broken after the CTB and HSI work on v.4.4.

I implemented an easy fix, tested by using the replay application...


aeoranday (approved):
LGTM. I also tested using the replay application and found no problem.

PR #292, opened 2024-04-03
Removed Separate Fragment Time Comparisons

aeoranday:
The check on TP ordering assumed that there was only one TP fragment per TimeSlice. Any TimeSlice that has multiple TP fragments then compares across fragments, and the last TP of the first is likely much later than the first TPs in the next fragments. This change resets the time comparison per fragment.

This was tested by running the replay application on TPStreams which contain multiple TP fragments per TimeSlice and weakly checking the resulting RawData file. The weak check was done by using the BundleN algorithms and assuring the bundle sizes were correct. There was not a cross comparison between the original TPStream and the resulting RawData.

### DIFF
--- a/plugins/TriggerPrimitiveMaker.cpp
+++ b/plugins/TriggerPrimitiveMaker.cpp
@@ -173,6 +173,7 @@ TriggerPrimitiveMaker::read_tpsets(std::string filename, int element)
 
     trgdataformats::TriggerPrimitive* tp_array = static_cast<trgdataformats::TriggerPrimitive*>(frag->get_data());
 
+    old_time_start = 0;
     for (size_t i(0); i < num_tps; i++) {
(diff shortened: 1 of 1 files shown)


ArturSztuc (approved):
LGTM.

Do you think this could play into the difference we see between number of TPs extracted from TA and number of TPs extracted from the buffer? Intuitively I think no, but you ran the tests

aeoranday:
No, I this is unrelated. Before this change, TriggerPrimitiveMaker would error at nanorc conf. It seems that we were using single fragment TimeSlices before and this went undetected (given the similar problem in `trgtools`). I'm unaware why there are multi-fragment TimeSlices now, but the fragments are unique so it doesn't seem like an issue.

We would also see problems on the TA-TC fragments.

PR #293, opened 2024-04-03
Glm/mlt

glehmannmiotto:
MLT restructuring (goes together with appdal schema changes)

### DIFF
--- a/plugins/ModuleLevelTrigger.cpp
+++ b/plugins/ModuleLevelTrigger.cpp
@@ -200,12 +200,13 @@ ModuleLevelTrigger::trigger_decisions_callback(dfmessages::TriggerDecision& deci
 {
     auto ts = decision.trigger_timestamp;
     auto tt = decision.trigger_type;
+    decision.run_number = m_run_number;
+    decision.trigger_number = m_last_trigger_number;
 
-    TLOG() << "Received decision " << decision.trigger_number << " timestamp "
+    TLOG() << "Received decision with timestamp "
              << decision.trigger_timestamp ;
     
     if ((!m_paused.load() && !m_dfo_is_busy.load())) {
-      decision.trigger_number = m_last_trigger_number;
       TLOG_DEBUG(1) << "Sending a decision with triggernumber " << decision.trigger_number << " timestamp "
              << decision.trigger_timestamp << " start " << decision.components.front().window_begin << " end " << decision.components.front().window_end
  	     << " number of links " << decision.components.size();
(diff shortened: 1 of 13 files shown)


plugins/ModuleLevelTrigger.cpp
```diff
@@ -342,444 +195,55 @@ ModuleLevelTrigger::do_resume(const nlohmann::json& /*resumeobj*/)
...
       }
 
     } else if (m_paused.load()) {
       ++m_td_paused_count;
-      m_td_paused_tc_count += pending_td.contributing_tcs.size();
-      TLOG_DEBUG(1) << "Triggers are paused. Not sending a TriggerDecision for pending TD with start/end times "
-                    << pending_td.readout_start << "/" << pending_td.readout_end;
+      //m_td_paused_tc_count += pending_td.contributing_tcs.size();
+      TLOG_DEBUG(1) << "Triggers are paused. Not sending a TriggerDecision for TD with timestamp and type "
+                    << ts << "/" << tt;
```
ArturSztuc, line 235: The type is a bitword (there can be multiple TCs of different types in TD), and I'm not sure if `auto` appreciates it, or just converts e.g. `000100010` (say, prescale & Timing types in this TD) to e.g. 34 (meaningless to reader).

Doesn't need to be resolved in this PR though, though maybe a TODO comment could be useful.

plugins/ModuleLevelTrigger.cpp
```diff
@@ -342,444 +195,55 @@ ModuleLevelTrigger::do_resume(const nlohmann::json& /*resumeobj*/)
...
     } else {
       ers::warning(TriggerInhibited(ERS_HERE, m_run_number));
-      TLOG_DEBUG(1) << "The DFO is busy. Not sending a TriggerDecision for candidate timestamp "
-                    << pending_td.contributing_tcs[m_earliest_tc_index].time_candidate;
+      TLOG_DEBUG(1) << "The DFO is busy. Not sending a TriggerDecision with timestamp and type "
+                    << ts << "/" << tt;
```
ArturSztuc, line 239: ditto

src/TCProcessor.cpp
```diff
@@ -0,0 +1,735 @@
...
+    if (tc_type == *it) {
+      ignore = true;
+      break;
+    }
+    ++it;
+  }
+  return ignore;
+}
+
+std::bitset<16>
```
ArturSztuc, line 470: Thinking out loud, all these 16s will need to be changed to 64s when merging v4 into develop.

I think v4 and v5 diverged so much, that maybe there's some motivation to NOT do a standard v4 merge into develop, and instead just re-implement whatever v4 features we had into v5 from scratch. It might be easier...

ArturSztuc (approved):
LGTM!

PR #295, opened 2024-04-10
Add support for CIB

nbarros:
This PR adds in the CIB as a new HSI-like object capable of generating Trigger Candidates. The code is largely based on the CTB code, with a few caveats for the possibility of having multiple insteances if the CIBModule to feed HSIEvents. 

By default the CIB is not enabled, and therefore this code should not affect previous operation. 

edit: Just as a note, ran the `minimal_system_quick_test.py` integration test with all the CIB related code and everything was fine.


bieryAtFnal (approved):
LGTM in so far as existing integtests continue to run successfully.

PR #299, opened 2024-04-16
Asztuc/tpwrapper fix

ArturSztuc:
This PR has two bug-fixes:
1. Correct the expected tick difference between TPs in trigger's TP buffer's `TPWrapper` to match the one in the readout's `TriggerPrimitiveAdapter`, fixing the missing-tps from the Trigger's TP buffer issue.
2. Fixing the replay application by adding the CIB line, needed after the CIB integration into the trigger.

Tests done:
2. Ran replay application with HMA algorithm before and after the fix: before the fix n TPs from buffer fragment is lower than in TA fragment, after the fix n TPs from buffer fragment is higher than in the TA fragment, as expected.
3. Ran offline on raw fragment asset with TPG using BundleN algorithm: the number of TPs is identical number of TPs from the readout buffer and from the trigger buffer fragments.
4. @aeoranday ran extra comparisons showing the discrepancy between the fragments from the two TP buffers (readout vs trigger) is fixed with this PR.

The very last TP from the readout window is always missing, however, from both buffers. This is because the SkipListLatencyBuffer (or maybe the Handler itself) is inclusive for the start of the readout window, and exclusive for the end of the readout window.

For comparison, here's the readouts `TriggerPrimitiveAdapter`: https://github.com/DUNE-DAQ/fdreadoutlibs/blob/production/v4/include/fdreadoutlibs/TriggerPrimitiveTypeAdapter.hpp.
We probably should be using the same code in the future.

### DIFF
--- a/plugins/TPBuffer.hpp
+++ b/plugins/TPBuffer.hpp
@@ -94,7 +94,7 @@ private:
     static const constexpr daqdataformats::SourceID::Subsystem subsystem = daqdataformats::SourceID::Subsystem::kTrigger;
     static const constexpr daqdataformats::FragmentType fragment_type = daqdataformats::FragmentType::kTriggerPrimitive;
     // No idea what this should really be set to
-    static const constexpr uint64_t expected_tick_difference = 1; // NOLINT(build/unsigned)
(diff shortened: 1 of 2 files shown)


aeoranday (approved):
LGTM. Though, I was a part of the testing and proving the improvement.

PR #306, opened 2024-06-03
Removed the TP buffers from the trigger

ArturSztuc:
Full explanation here: https://github.com/DUNE-DAQ/daqconf/pull/462


wesketchum (approved):
LGTM

PR #310, opened 2024-06-13
Unique and short thread names for trigger's DAQModules

ArturSztuc:
All the trigger `DAQModule`s `do_work` threads now have unique names, taken from the module's unique name -- which is in turn set in the `daqconf`.  At the same time, some of the modules unique names had to be renamed as they were more than 16 characters long -- and they must be 16 chars or less, as explained [HERE](https://linux.die.net/man/3/pthread_setname_np).
This doesn't change how the trigger works in any way, but will enable us to pin threads to specific CPUs if we require this for NP04. 

This PR goes together with: 

Tested by:
1. Running the he integration test `3ru_3df_multirun_test.py`.
2. Successfully generating a configuration with `--debug` to generate .dot files and looking through them.
3. Running offline with 3 readout units, and seeing triggers as expected.

trigger

### DIFF
--- a/include/trigger/Tee.hxx
+++ b/include/trigger/Tee.hxx
@@ -59,7 +59,7 @@ template<class T>
 void
 Tee<T>::do_start(const nlohmann::json&)
 {
-  m_thread.start_working_thread("tee");
+  m_thread.start_working_thread(get_name());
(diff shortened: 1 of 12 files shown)


MRiganSUSX (approved):
These changes look great. 

There are some places where a specific name is assumed (for example `mlt` in fragment_producers) but this is not affecting that (as the name is kept the same).

PR #311, opened 2024-06-13
Extra operational monitoring in the trigger

ArturSztuc:
Extra operational monitoring for the trigger's buffers (both TA and TC), and TAZipper (added occupancy monitoring). Still need to update grafana.

Tested by:
1. Running the the integration test `3ru_3df_multirun_test.py`.
2. Running with monitoring enabled, and scanning through the generated info file for the new monitoring information.


aeoranday (approved):
LGTM. Tested using the `daqsystemtest` bundle and reading the generated trigger info file. All new monitoring values were present and nothing appeared outstanding.

PR #312, opened 2024-06-14
extending TC types with new CTB bits

MRiganSUSX:
Correlates with https://github.com/DUNE-DAQ/trgdataformats/pull/30
New CTB related types and their mapping for CTBTriggerCandidateMaker.


ArturSztuc (approved):
LGTM

PR #314, opened 2024-06-17
Bugfix for the TA & TC adapters in the trigger

ArturSztuc:
Fix to the expected width of the trigger candidate & trigger activity for the TA/TC adapter (when placing in the latency buffer).

Tested by running online with eight CTB bits on: `CTBTriggerCandidateMaker` would create eight TCs with identical `start_time`. MLT would merge this into one TD, so in the trigger-words from the raw dat we would see all eight for each TD, but only one TC being retrieved from the trigger. This is no longer the case, now we have all 8 TCs being retrieved from the latencybuffer.

### DIFF
--- a/plugins/TABuffer.hpp
+++ b/plugins/TABuffer.hpp
@@ -68,6 +68,7 @@ private:
     bool operator<(const TAWrapper& other) const
     {
       return this->activity.time_start < other.activity.time_start;
+      return std::tie(this->activity.time_start, this->activity.channel_start) < std::tie(other.activity.time_start, other.activity.channel_start);
     }
 
     uint64_t get_first_timestamp() const // NOLINT(build/unsigned)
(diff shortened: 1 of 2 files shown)


MRiganSUSX (approved):
I have tested that this resolves the issue that the CTB group has reported.

The TC buffers now 'produce' the expected number of TCs when requested. 

HDF5 DUMP:
```
	Trigger_Candidate fragment with SourceID Trigger_0x00000000 from subdetector DAQ has size = 408 -----
		Readout window before = 1000, after = 1000
		Number of TCs in this fragment=6, overall number of referenced TAs=0, size of TC data=56
		First TC type = kCTBFakeTrigger (13), TC algorithm = 2, number of TAs = 0
		First TC start time=79554165817071816, end time=79554165817073816, and candidate time=79554165817072816
```
It is also possible to parse the TC types:
```
Path: /TriggerRecord00001.0000/RawData/Trigger_0x00000000_Trigger_Candidate
Trig num: 1
This frag TCs:  6
0
336
Type.kCTBFakeTrigger
1
280
Type.kCTBBeam
2
224
Type.kCTBBeamHiLoPressChkv
3
168
Type.kCTBBeamLoPressChkv
4
112
Type.kCTBOffSpillCosmic
5
56
Type.kCTBCosmic
```

Also passes the integtests (minimal, fake data producer, 3ru 1df)

PR #319, opened 2024-07-04
Custom subdetector readout for MLT

ArturSztuc:
With this PR the user can add custom subdetector readout window map for the MLT, which will set the readout window for specific subdetector -- e.g. 5ms for the PDS, but 3ms for the TPC etc.

This PR goes together with PRs:  https://github.com/DUNE-DAQ/daqconf/pull/482, https://github.com/DUNE-DAQ/fddaqconf/pull/45

In the trigger configuration, simply add something along:
```
"trigger": {
    "mlt_detector_readout_map": [
        {"subdetector": "HD_PDS", "time_before": 150000, "time_after": 150000},
        {"subdetector": "HD_CRT", "time_before": 50000, "time_after": 100000},
    ]
}
```
The above will result in the PDS readout always being 300k ticks wide, CRT always being 150k ticks wide, and all the rest (TPC, software triggers etc) will have their defaults.

This only works for the raw detector readout buffers, not e.g. trigger buffers (so effectively, only for `SourceID` with subsystem of `kDetectorReadout`).

The names of the available subdetectors are all defined inside of `detdataformats::DetID` [HERE](https://github.com/DUNE-DAQ/detdataformats/blob/11a10360c1b16fb1596793253ec90dd61bdcef95/include/detdataformats/DetID.hpp#L39-L53) and [HERE](https://github.com/DUNE-DAQ/detdataformats/blob/11a10360c1b16fb1596793253ec90dd61bdcef95/include/detdataformats/detail/DetID.hxx#L66-L95), if the user supplies wrong subdetector name the trigger application will crash at conf time with an appropriate error message.

Tests:
1. `3ru_3df_multirun_test.py` integration fully passes.
2. Offline test with custom detector map for `HD_TPC` on an asset file.
The written WIBeth fragments were of that exact window, whereas all the other fragments had windows set by the TCs (prescale trigger algorithm).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,7 @@ daq_codegen(
   tasetsink.jsonnet
   tpchannelfilter.jsonnet
   cibtriggercandidatemaker.jsonnet
+  DEP_PKGS hdf5libs
   TEMPLATES Structs.hpp.j2 Nljs.hpp.j2 )
 
 daq_codegen(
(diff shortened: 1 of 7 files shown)


wesketchum (approved):
looks ok to me, and is working at EHN1. thanks!

PR #328, opened 2024-07-22
Adding latency monitoring for missing templates in TriggerGenericMaker

MRiganSUSX:
This relates to https://github.com/DUNE-DAQ/trigger/pull/327 which makes use of previously unused template in TGM. The common latency monitoring in the TGM was extended to cover this case as well.

Tested both offline and online (np04).


ArturSztuc (approved):
LGTM

Passes test with `3ru_3df_multirun_test.py`, including branches from https://github.com/DUNE-DAQ/trigger/pull/327.
This was all also successfully tested on PD2HD, as described in https://github.com/DUNE-DAQ/trigger/pull/327.

PR #331, opened 2024-07-31
Bugfixing the TAWrapper and TCWrapper

ArturSztuc:
Bugfixing the TA and TC wrapper objects that are needed to insert TAs and TCs into the latency buffers.

First bugfix is the same as in v4 https://github.com/DUNE-DAQ/trigger/pull/314.
The TC and TA objects need to have expected tick-difference between adjacent TAs to be 1, rather than 16 -- because:
1. TA/TC size differs from object to object 
2. Buffer can contain TAs from different parallel streams, meaning there can be 2 TAs in the buffer that have the same timestamp (coming from different APAs, planes, algorithms etc).

Second bugfix is specific to v5, the object's `begin()` function needs to start at the beginning of the serialised vector of bits. `TAWrapper`s and `TCWrapper`s store both the object itself, and it's serialised form (should change this!), but buffers, dfo etc. only read the latter.

Tested by running the `test-session`, and using `HDF5LIBS_TestDumpRecord` on the output raw data files.
Before the bugfix the "number of TAs in TC" was nonsense.
After the bugfix it was always 1 for the prescale algorithm, which is the expected number.

### DIFF
--- a/include/trigger/TAWrapper.hpp
+++ b/include/trigger/TAWrapper.hpp
@@ -41,7 +41,7 @@ namespace trigger {
     // comparable based on first timestamp
     bool operator<(const TAWrapper& other) const
     {
-      return this->activity.time_start < other.activity.time_start;
+      return std::tie(this->activity.time_start, this->activity.channel_start) < std::tie(other.activity.time_start, other.activity.channel_start);
     }
 
     uint64_t get_first_timestamp() const // NOLINT(build/unsigned)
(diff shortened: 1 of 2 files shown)


glehmannmiotto (approved):
Those fixes look good, but I still think that it is not correct to keep in the wrappers both the Activity and the overlay buffer. If the overlay is what is needed, then the TriggerActivity should not need to be also kept in the struct.
For future discussion....

PR #332, opened 2024-08-06
Updating trigger HSISourceModel with v4 changes

ArturSztuc:
This PR needs to go together with https://github.com/DUNE-DAQ/appmodel/pull/99.

`HSISourceModel` now has configurable output TC-types, can deal with scenarios where there are multiple signal bits switched on in `HSIEvent` simultaneously, and has an option for prescaling.

This can be also considered as "initial port" of CTB TCM and CIB TCM on the trigger side: since `HSISourceModel` can now output any TC-type, we can configure e.g. a separate `HSIEventToTCApplication` application, with input connection to the CTBModule (when written), and `HSIEvent` bits configured to output specific CTB TC types. The same goes for CIB.

### DIFF
--- a/src/trigger/HSISourceModel.hpp
+++ b/src/trigger/HSISourceModel.hpp
@@ -26,6 +26,16 @@
 
 namespace dunedaq::trigger {
 
+/**
+ * @struct HSISignal
+ * @brief Struct holding configuration for one HSI signal bit
+ */
+struct HSISignal
+{
+  triggeralgs::TriggerCandidate::Type type;
+  triggeralgs::timestamp_t time_before;
+  triggeralgs::timestamp_t time_after;
+};
 
 class HSISourceModel : public datahandlinglibs::SourceConcept
 {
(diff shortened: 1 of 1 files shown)


ArturSztuc:
Forgot to mention. Tested with drunc:
Created a hsi configuration that sends `HSIEvent` with the first and second bits switched on.
At the same time, configured `HSISourceModel` to convert the first bit into `kTiming` and second to `kMichelElectron`, with different readout windows.
Ran with drunc, and confirmed with `HDF5LIBS_TestDumpRecord` and grafana that both trigger types are present in the trigger records.
Confirmed prescale working by changing prescale and monitoring trigger rates on grafana.

ArturSztuc:
@MRiganSUSX @schhibra could someone please review this PR? It's been a week, and I would really like to merge it before doing more dev on v5.
There's also a triggeralgs PR in appmodel [HERE](https://github.com/DUNE-DAQ/appmodel/pull/100).

glehmannmiotto (approved):
LGTM; beware that there is also the opmon branch pending to be merged; that one should get in first in order not to break the next nightly.

PR #334, opened 2024-08-15
Renaming two CTB TC types

ArturSztuc:
Renaming two CTB TC types.
Standard integration tests pass. This PR needs to go together with https://github.com/DUNE-DAQ/trgdataformats/pull/39


MRiganSUSX (approved):
lgtm

just a note that this will have impact on backwards compatibility.

PR #336, opened 2024-08-22
CIB fixes

nbarros:
This PR (together with corresponding PRs on cibmodules and fddaqconf brings fixes to generate TCs from the CIB/IoLaser system.


MRiganSUSX (approved):
lgtm
(seems to be only log and formatting changes)

PR #342, opened 2024-09-13
Trigger latencies + monitoring (v5)

MRiganSUSX:
This PR implements monitoring of latencies across trigger. 
Requires appmodel PR: https://github.com/DUNE-DAQ/appmodel/pull/126.

Changes:
- new latency class that is used throughout: simple, stores and retrieves timestamps. Can be configured to use ms or us. 
- currently implemented in: CTCM, RTCM, MLT, TXProcessors, HSISM
- also publishes latency information to opmon (all above), using shared proto message for easy changes. Grafana plots are already in place.
- 'weird' case for MLT where two instances are used, one is for readout windows (as was the case in v4)

For appmodel changes:
- created LatencyMonitoringCong class, simply allowing to enable/disable latency monitoring
- instantiated object using this class
- expanded relevant classes and objects that are using this. Tried to abstract the configuration to limit copies of the same code (ie made part of DataProcessor), but not sure it's the best it could have been.

other small changes:
- removed tlog block in DHM (left in by mistake)
- some consistency changes for opmon variables


include/trigger/Latency.hpp
```diff
@@ -0,0 +1,102 @@
...
+      } else {
+        return 0;
+      }
+    }
+
+  private:
+    std::atomic<uint64_t> m_latency_in;  // Member variable to store latency_in
+    std::atomic<uint64_t> m_latency_out; // Member variable to store latency_out
+    std::atomic<double> m_clock_ticks_conversion; // Dynamically adjusted conversion factor for clock ticks
+    mutable std::mutex m_mutex;
```
mroda88: This class implementation is a bit weird from the technical point of view. 
The mutex requirement is unclear because you already have atomic counters. 

If you need we can discuss it.

schema/trigger/opmon/latency_info.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package dunedaq.trigger.opmon;
+
+// Message for latency variables
+// Latency represents the difference between current system (clock) time and the data time of particular (TX) data object
+// Units are ms
```
mroda88: I would recommend that we publish microseconds.

mroda88 (commented):
From the point of view of what information is published, this seems ok. 

The usage of the locking is doubtful at the moment.

schema/trigger/opmon/moduleleveltrigger_info.proto
```diff
@@ -24,3 +24,11 @@ message TriggerDecisionInfo {
   uint32 paused = 4;          // Number of paused (triggers are paused)
   uint32 inhibited = 5;       // Number of inhibited (DFO is busy)
 }
+
+// Message for MLT TD requests latency vars
+// Latency represents the difference between current system (clock) time and the requested TD readout window (start/end)
+// Units are ms
```
mroda88: I would recomment to publish us

plugins/CustomTCMaker.cpp
```diff
@@ -107,6 +107,15 @@ CustomTCMaker::generate_opmon_data()
   info.set_tc_failed_sent_count( m_tc_failed_sent_count.load() );  
 
   this->publish(std::move(info));
+
+  if ( m_running_flag.load() && m_latency_monitoring.load() ) {
```
ArturSztuc:
```suggestion
  if ( m_latency_monitoring.load() && m_running_flag.load() ) { 
```
Not an important comment, I'm only writing this because I've learned something new a few weeks back and want to share :) Unnecessary micro-optimisation, but in C++ order inside if statement can matter of operations like `&&`, `||` (not true for all operators).
So in your case, whilst we're running, we will always load 2 booleans, even if we don't want monitoring. With the suggestion above, if monitoring is off, we will only load 1 boolean. Matters more task-heavy workflows so probably not here, but maybe in e.g. `TPDataProcessor` where every little counts.

plugins/CustomTCMaker.cpp
```diff
@@ -125,13 +134,20 @@ CustomTCMaker::do_configure(const nlohmann::json& /*obj*/)
   //// This parameter controls how many new timestamps are calculated when needed
   //// Currently precalculates events for the next 60 seconds
   //m_sorting_size_limit = 60 * m_conf0>clock_frequency_hz;
+  
+  m_latency_monitoring.store( m_conf->get_latency_monitoring_conf()->get_enable_latency_monitoring() );
```
ArturSztuc: `do_configure` is deprecated in v5.
All the config should be inside of `void init(std::shared_ptr<appfwk::ModuleConfiguration> mcfg);`

Technically there's also a new `conf`, but only in non-`DAQModule` classes I think.

plugins/CustomTCMaker.cpp
```diff
@@ -107,6 +107,15 @@ CustomTCMaker::generate_opmon_data()
   info.set_tc_failed_sent_count( m_tc_failed_sent_count.load() );  
 
   this->publish(std::move(info));
+
+  if ( m_running_flag.load() && m_latency_monitoring.load() ) { 
+    opmon::TriggerLatency lat_info;
+
+    lat_info.set_latency_in( m_latency_instance.get_latency_in() );
+    lat_info.set_latency_out( m_latency_instance.get_latency_out() );
```
ArturSztuc: I think there's a logic issue here:
`get_latency_in()` function seems to compare the object time-stamp with the current time. That's a number we want to obtain e.g. right after obtaining a new object, right after creating an object, or right before sending the object elsewhere. 

But here we're comparing it against time at opmon measurement time instead, which will add extra latency (because this opmon measurement might happen e.g. after the object was already sent, so we're comparing objects timestamp against system time after the object is already gone from the system).

So in other words, `get_current_system_time()` should be called when we take the latency measurement, not when we're sending it to opmon after the measurement.

plugins/MLTModule.cpp
```diff
@@ -106,6 +109,21 @@ MLTModule::generate_opmon_data()
...
+    // TC in, TD out
+    opmon::TriggerLatency lat_info;
+    lat_info.set_latency_in( m_latency_instance.get_latency_in() );
+    lat_info.set_latency_out( m_latency_instance.get_latency_out() );
+    this->publish(std::move(lat_info));
+
+    // vs readout window requests
+    opmon::ModuleLevelTriggerRequestLatency lat_request_info;
+    lat_request_info.set_latency_window_start( m_latency_requests_instance.get_latency_in() );
+    lat_request_info.set_latency_window_end( m_latency_requests_instance.get_latency_out() );
```
ArturSztuc, line 124: Ditto re. getting the system time

plugins/RandomTCMakerModule.cpp
```diff
@@ -76,12 +76,22 @@ RandomTCMakerModule::generate_opmon_data()
   info.set_tc_failed_sent_count( m_tc_failed_sent_count.load() );
 
   this->publish(std::move(info));
+
+  if ( m_running_flag.load() && m_latency_monitoring.load() ) {
+    opmon::TriggerLatency lat_info;
+
+    lat_info.set_latency_in( m_latency_instance.get_latency_in() );
+    lat_info.set_latency_out( m_latency_instance.get_latency_out() );
```
ArturSztuc: Ditto re. getting the system-time at measurment time vs at sending-to-opmon time

plugins/RandomTCMakerModule.cpp
```diff
@@ -76,12 +76,22 @@ RandomTCMakerModule::generate_opmon_data()
...
+
+    this->publish(std::move(lat_info));
+  }
 }
 
 void
 RandomTCMakerModule::do_configure(const nlohmann::json& /*obj*/)
 {
   //m_conf = obj.get<randomtriggercandidatemaker::Conf>();
+  m_latency_monitoring.store( m_conf->get_latency_monitoring_conf()->get_enable_latency_monitoring() );
```
ArturSztuc: Ditto re. `do_configure` being deprecated in v5, should move to init.

src/TAProcessor.cpp
```diff
@@ -108,6 +113,15 @@ TAProcessor::generate_opmon_data()
   info.set_tc_failed_sent_count( m_tc_failed_sent_count.load() );
 
   this->publish(std::move(info));
+
+  if ( m_running_flag.load() && m_latency_monitoring.load() ) {
+    opmon::TriggerLatency lat_info;
+
+    lat_info.set_latency_in( m_latency_instance.get_latency_in() );
+    lat_info.set_latency_out( m_latency_instance.get_latency_out() );
```
ArturSztuc, line 121: Ditto re. getting system time at measurement time vs at sending-to-opmon time

src/TCProcessor.cpp
```diff
@@ -194,6 +194,15 @@ TCProcessor::generate_opmon_data()
   info.set_tds_cleared_tc_count( m_tds_cleared_tc_count.load() );
 
   this->publish(std::move(info));
+
+  if ( m_running_flag.load() && m_latency_monitoring.load() ) {
+    opmon::TriggerLatency lat_info;
+
+    lat_info.set_latency_in( m_latency_instance.get_latency_in() );
+    lat_info.set_latency_out( m_latency_instance.get_latency_out() );
```
ArturSztuc, line 209: Ditto re. getting system time at measurement time vs at sending-to-opmon time

src/TPProcessor.cpp
```diff
@@ -111,6 +115,15 @@ TPProcessor::generate_opmon_data()
   info.set_ta_failed_sent_count( m_ta_failed_sent_count.load() );
 
   this->publish(std::move(info));
+
+  if (m_running_flag.load() && m_latency_monitoring.load() ) {
+    opmon::TriggerLatency lat_info;
+
+    lat_info.set_latency_in( m_latency_instance.get_latency_in() );
+    lat_info.set_latency_out( m_latency_instance.get_latency_out() );
```
ArturSztuc, line 123: Ditto re. getting system time at measurement time vs at sending-to-opmon time

src/trigger/HSISourceModel.hpp
```diff
@@ -173,6 +184,15 @@ class HSISourceModel : public datahandlinglibs::SourceConcept
     info.set_tcs_dropped_count( m_tcs_dropped_count );
 
     this->publish(std::move(info));
+
+    if ( m_running_flag.load() && m_latency_monitoring.load() ) {
+      opmon::TriggerLatency lat_info;
+
+      lat_info.set_latency_in( m_latency_instance.get_latency_in() );
+      lat_info.set_latency_out( m_latency_instance.get_latency_out() );
```
ArturSztuc, line 191: Ditto re. getting system time at measurement time vs at sending-to-opmon time

ArturSztuc (changes requested):
Overall looks very good! There are only 2 serious comments:

1. `do_configure` is deprecated in v5, and we need to use `init` instead (or `conf`, but only in non-`DAQModule` classes).
2. There's a logic issue with when we're loading the "current system time". We should do it at latency measurement time, like we did in v4, not at sending-to-opmon time. This will add unnecessary latency, especially in slow-flow system (e.g. 1hz rates)

MRiganSUSX:
Thanks @ArturSztuc and @mroda88 for insights. 

Here is v2, changes:
In trigger:
- reworked the latency class (again), to be simpler, and modular
- optimized the if conditions check as Artur suggested
- moved config loading to init or conf as requested
- the default for latency instances is now us (microseconds) throughout (and plots have been updated to reflect this, thanks to @mroda88)

In appmodel:
- previous PR was closed (https://github.com/DUNE-DAQ/appmodel/pull/126)
- new PR: https://github.com/DUNE-DAQ/appmodel/pull/141
- this is using attributes only to pass on the latency boolean

This was tested:
- offline with example configs with additional debug logs
- using the ehn1 example sessions and monitoring plots at grafana
- using `dbt-unittest-summary.sh`
- using integration tests: minimal_system_quick_test, tpstream_writing_test, 3ru_1df_multirun_test

Some issues and potential improvements are mentioned https://github.com/DUNE-DAQ/trigger/issues/347, for future consideration.

plugins/CustomTCMaker.cpp
```diff
@@ -236,11 +251,13 @@ CustomTCMaker::send_trigger_candidates()
     }
 
     triggeralgs::TriggerCandidate candidate = create_candidate(m_next_trigger_timestamp, m_tc_timestamps.front().first);
+    if (m_latency_monitoring.load()) m_latency_instance.update_latency_in( candidate.time_candidate );
     m_tc_made_count++;
 
     TLOG_DEBUG(1) << get_name() << " at timestamp " << m_timestamp_estimator->get_timestamp_estimate()
                   << ", pushing a candidate with timestamp " << candidate.time_candidate;
 
+    if (m_latency_monitoring.load()) m_latency_instance.update_latency_out( candidate.time_candidate );
```
ArturSztuc, line 258: Do we need both latency in and latency in and out here? The time difference here is literally the time taken to do `m_tc_made_count++`, which will be in nanoseconds. Out latency would be more than enough.

I think in & out makes sense if we have some input data, a processing stage, and output data. In the standalone TC makers we just have the output data

plugins/RandomTCMakerModule.cpp
```diff
@@ -208,11 +223,14 @@ RandomTCMakerModule::send_trigger_candidates()
...
     next_trigger_timestamp = m_timestamp_estimator->get_timestamp_estimate();
     triggeralgs::TriggerCandidate candidate = create_candidate(next_trigger_timestamp);
+
+    if (m_latency_monitoring.load()) m_latency_instance.update_latency_in( candidate.time_candidate );
     m_tc_made_count++;
 
     TLOG_DEBUG(1) << get_name() << " at timestamp " << m_timestamp_estimator->get_timestamp_estimate()
                   << ", pushing a candidate with timestamp " << candidate.time_candidate;
 
+    if (m_latency_monitoring.load()) m_latency_instance.update_latency_out( candidate.time_candidate );
```
ArturSztuc, line 231: Ditto here, we only need one latency in standalone makers, having in and out doesn't make sense, and will give identical numbers (unless we switch to nanoseconds or picoseconds.

ArturSztuc (approved):
LGTM!

The only further comment I have, and something I didn't notice earlier, is that the most of the `StandaloneTCMakers` don't need both latency in and latency out. They create a new object and then send it straight away, so the time difference between these two is in nanoseconds. Latency out at sending would be enough (only for Random & Custom TC maker, HSI of course does receive an input and processes it).

I don't mind if that's in this PR or a separate one.

MRiganSUSX:
Thanks Artur,
I updated the standalone makers to only report latency out now. This unfortunately needed a different opmon message, so that we are not sending a message that is half empty.

ArturSztuc (approved):
LGTM, thanks for all the work!

PR #345, opened 2024-10-02
Modify TCMakers so that the TriggerCandidates they create can be sent

eflumerf:
through the DataSubscriberModule data path.


plasorak (approved):
Still getting TD on the other end :) LGTM

PR #346, opened 2024-10-08
Fixing TCProcessor's high mlt-dec thread occupancy

ArturSztuc:
`mlt-dec` thread had 100% occupancy. This is reminiscent of the v4 mlt, that also always had 100% thread occupancy. 
In both cases it's because of a while loop without any `sleep_for`, so it runs at 100% non-stop. 

This fix adds a check to the TD-sending function inside the while loop, that makes the thread `sleep_for` 10ms (and therefore releasing this thread to other tasks) if there's no TDs pending.

Tested by running through drunc, checking if the data is still being saved, and `mlt-dec` now has orders of magnitude lower occupancy.

Potentially adds up to 10ms latency to the TDs being sent. Can reduce if considered too high.

### DIFF
--- a/src/TCProcessor.cpp
+++ b/src/TCProcessor.cpp
@@ -68,6 +68,14 @@ TCProcessor::stop(const nlohmann::json& args)
 {
   inherited::stop(args);
   m_running_flag.store(false);
+
+  // Make sure condition_variable knows we flipped running flag
+  {
+    std::lock_guard<std::mutex> lock(m_td_vector_mutex);
+    m_cv.notify_all();
+  }
+
+  // Wait for the TD-sending thread to stop
   m_send_trigger_decisions_thread.join();
 
   // Drop all TDs in vectors at run stage change. Have to do this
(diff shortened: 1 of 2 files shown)


eflumerf:
This seems like an ideal use-case for condition variables: https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for

ArturSztuc:
Hi @eflumerf, thank you for this suggestion, and I agree that it makes much more sense to use it!
I adapted `condition_variable::wait_for` for now. I tested it and it works, so the commit is already here ready for review.

I tried to adapt straight up `condition_variable::wait` instead, which I think would be even better... it works, but causes MLT to crash on `stop`, which I couldn't resolve (I tried sending signal to the conditional variable at stop time, making conditional variable to also depend on the `m_running_flag` etc).

So for now I'll stick with `condition_variable::wait_for`, and switch to `condition_variable::wait` in the future when I resolve this.

ArturSztuc:
PR ready to review.
@eflumerf sorry for the noise. I got `condition_variable::wait` to fully work, so I scrapped the `condition_variable::wait_for` implementation.

Tested by running through drunc, looking at saved TCs in the output raw file, looking at htop for `mlt-dec`.

bieryAtFnal (approved):
I've confirmed that the mlt-dec thread no longer uses a noticeable amount of CPU when running the minimal_system_quick_test.

eflumerf (approved):
Builds, unit tests pass, minimal_system_quick_test passes. Merging so it gets into the nightly

PR #348, opened 2024-10-14
Asztuc/mlt v4 to v5

ArturSztuc:
This PR has to go together with https://github.com/DUNE-DAQ/appmodel/pull/143

Porting MLT changes from v4 to v5:
* Ignoring overlaps between TCs, to deal with pileups in PD2
* Configurable readout windows for different sub-detectors, enabling wider readout window for the PDS than TPC

Tested by running offline with custom readout map expanding TPC (but not TA/TP/Trigger data), and with high rates but ignoring overlaps.

### DIFF
--- a/plugins/MLTModule.cpp
+++ b/plugins/MLTModule.cpp
@@ -294,8 +294,16 @@ MLTModule::trigger_decisions_callback(dfmessages::TriggerDecision& decision )
 
     // Overwrite the component's readout window if we have custom
     // subdetector--readout window map
-    for ( const auto& [sourceid, window] : m_subdetector_readout_window_map ) {
+    for ( const auto& [subdetectorid, window] : m_subdetector_readout_window_map ) {
       for (auto& request: decision.components) {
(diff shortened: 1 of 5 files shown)


eflumerf (approved):
Builds, unit test passes, regression tests pass

plugins/MLTModule.cpp
```diff
@@ -44,6 +44,28 @@ MLTModule::MLTModule(const std::string& name)
   // clang-format on
 }
 
+
+std::map<std::string, int>
+MLTModule::decode_geoid(uint64_t _geoid_int)
```
eflumerf, line 49: It may have gone away with v5, but I thought we had GeoID decoding utilities somewhere else...

ArturSztuc: The functions I used in v4 for that seemed to have disappeared in v5. 

Moreso, GeoID is now encoded into one long integer, rather than being an object. I found two separate (and identical) "encode" functions to make these, but no "decoding" function, which is what I need here... 
It definitely would be good to have them somewhere more reasonable than in the trigger.

PR #350, opened 2024-10-16
RandomTCMaker overhaul

ArturSztuc:
This PR needs to be merged together with:
1. https://github.com/DUNE-DAQ/appmodel/pull/146
2. https://github.com/DUNE-DAQ/daqsystemtest/pull/132
3. https://github.com/DUNE-DAQ/drunc/pull/261

A small overhaul of `RandomTCMaker`:
1. The maker now accepts `trigger_rate_hz` rather than `trigger_interval_ticks`
5. New function `do_change_trigger_rate` that will be triggered when user enters `change-rate` in `drunc` live session to change the trigger rate live.
6. More configurable: the TC maker now accepts `TCReadoutMap` configuration object, where user can specify the TC window (time before and time after) as well as the output TC type (in integer for now, will be in string later, e.g. kRandom, kTiming, kSupernova etc).
7. The clock speed is now pulled directly from the "session", rather than having to be manually added in the configuration.

Tested by:
1. Running with current `drunc` automatically sets the trigger rate to 1 as expected given the current drunc dev.
2. Running with https://github.com/DUNE-DAQ/drunc/pull/261 doesn't override the trigger rate at `do_start`.
3. Can change trigger rate live with https://github.com/DUNE-DAQ/drunc/pull/261 and https://github.com/DUNE-DAQ/daqsystemtest/pull/127
4. Checked the output raw data file to make sure the trigger rates were as expected for all the runs above.
8. Changed the output TC type to `kSupernova` to see if the new config works.

### DIFF
--- a/plugins/RandomTCMakerModule.cpp
+++ b/plugins/RandomTCMakerModule.cpp
@@ -44,6 +44,7 @@ RandomTCMakerModule::RandomTCMakerModule(const std::string& name)
   register_command("start", &RandomTCMakerModule::do_start);
   register_command("stop_trigger_sources", &RandomTCMakerModule::do_stop);
   register_command("scrap", &RandomTCMakerModule::do_scrap);
+  register_command("change_rate", &RandomTCMakerModule::do_change_trigger_rate);
 }
 
 void
@@ -61,7 +62,17 @@ RandomTCMakerModule::init(std::shared_ptr<appfwk::ModuleConfiguration> mcfg)
      m_time_sync_source = get_iom_receiver<dfmessages::TimeSync>(con->UID());
   }
   m_conf = mtrg->get_configuration();
+
+  // Get the TC out configuration
+  m_tc_readout = m_conf->get_tc_readout();
+
   m_latency_monitoring.store( m_conf->get_latency_monitoring() );
+
+  // Get the clock speed from detector configuration
+  m_clock_speed_hz = mcfg->configuration_manager()->session()->get_detector_configuration()->get_clock_speed_hz();
+  m_trigger_rate_hz = m_conf->get_trigger_rate_hz();
+  TLOG() << "Clock speed is: " << m_clock_speed_hz;
+  TLOG() << "Output trigger rate is: " << m_trigger_rate_hz;
 }
 
 void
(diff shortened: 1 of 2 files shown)


plasorak:
Here is what I did from today nightly `NFD_DEV_241016_A9`:
 - merged `develop` into your `appmodel`, `daqsystemtest`, and `trigger` branches
 - merged `plasorak/trigger-rate` into your `daqsystemtest`.
 - merged `develop` into the `plasorak/trigger-rate` of `drunc`.

When I do that, and run the local-1x1-config, and decide to enter a trigger rate of 0, I am not sure I get the correct behaviour:
 - raw datafile increasing very quickly
 - 1.6GB worth of logs for the MLT (full of ```
2024-Oct-16 19:29:32,626 LOG [void dunedaq::trigger::MLTModule::trigger_decisions_callback(dunedaq::dfmessages::TriggerDecision&) at /nfs/home/plasorak/NFD_DEV_241016_A9_trig/sourcecode/trigger/plugins/MLTModule.cpp:311] Received decision with timestamp 108068735763153182
```
 - the logs of tc maker full of ```
2024-Oct-16 19:29:25,193 LOG [std::string dunedaq::iomanager::get_uri_for_connection(const dunedaq::confmodel::NetworkConnection*) at /tmp/root/spack-stage/spack-stage-iomanager-NB_DEV_241016_A9-gq4tjwdkh7pk6rjs2bohcqtcb26vwxzo/spack-src/include/iomanager/SchemaUtils.hpp:94] Getting URI for network connection tas_.*
```

ArturSztuc:
Thanks @plasorak , I resolved this for the RTCM, but HSI modules still output lots of timing events with trigger-rate=0. That's outside of the trigger through... @strilov?

In RTCM it was because trigger rate is divided from the clock speed to calculate "time to the next TC". When trigger rate is 0, we divide by 0, creating NULL and freaking this whole thing out.

plasorak:
All right, minimal system quick test, long readout window test, 3ru1df test, 3ru3df test and tpstream writer test all pass (baring a couple of anomalous WIBEth fragment sizes in 3ru1df). Merging.

PR #351, opened 2024-10-21
Reworking TC-type configs

ArturSztuc:
This PR has to go together with:
1. https://github.com/DUNE-DAQ/appmodel/pull/150
2. https://github.com/DUNE-DAQ/daqsystemtest/pull/136

Done:
1. Reworked `TCProcessor` and `RandomTCMaker` configurations to use string-based TC-type inputs, rather than integer-based.
2. Removed `use_readout_map` and `use_roi_readout` from configuration inputs: now assumed to be `True` if readout map or roi map is provided.
3. Changed Trigger configuration error message.

Tests:
1. Compiles.
2. `pytest -s ${DAQSYSTEMTEST_SHARE}/integtest/3ru_3df_multirun_test.py` runs
3. Changed TC-types in the configurations, ran a test session, and confirmed correct outputs with `HDF5LIBS_TestDumpRecord`.

### DIFF
--- a/include/trigger/Issues.hpp
+++ b/include/trigger/Issues.hpp
@@ -28,7 +28,7 @@
 
 namespace dunedaq {
 
-ERS_DECLARE_ISSUE(trigger, InvalidConfiguration, "An invalid configuration object was received", ERS_EMPTY)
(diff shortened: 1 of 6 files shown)


aeoranday (approved):
LGTM. I tested using `daqsystemtest` and local sessions with varying configurations. Local sessions were confirmed using `trgtools`.

PR #356, opened 2024-11-05
Bugfixing RTCM for v5.2

ArturSztuc:
Bugfixing `RandomTCMaker` after v5.2 performance tests. 

**Bug**: when running with trigger rate of 0.01Hz or lower, RTCM would spit out TCs at much higher rate.
**Fix**: The issue was with using `int` for time intervals rather than `uint64_t`s. Changed from `int` to `uint64_t`
**Test**: Ran test configuration with low trigger rate (0.01Hz) on code before and after the fix for 300s, comparing raw file before and after. Changed from 1000s of triggers to 3, as expected.

### DIFF
--- a/plugins/RandomTCMakerModule.cpp
+++ b/plugins/RandomTCMakerModule.cpp
@@ -198,19 +198,19 @@ RandomTCMakerModule::create_candidate(dfmessages::timestamp_t timestamp)
   return candidate;
 }
 
-int
+uint64_t
 RandomTCMakerModule::get_interval(std::mt19937& gen)
 {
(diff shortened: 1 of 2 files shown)


bieryAtFnal (approved):
LGTM.
I used the `change_rate` command to set the trigger rate to 0.005 (one per 200 seconds).  In a 620-second run, I observed four triggers, which is in the range of what we would expect (one at 0, 200, 400, and 600 seconds into the run).

eflumerf (approved):
Observed using change_rate_test

PR #358, opened 2024-11-05
Add a test of the change-rate command to RandomTriggerCandidateMaker

eflumerf:


ArturSztuc (approved):
Overall looks good to me, although the unit test doesn't pass for me.
I either keep getting:

```🚨 Event count 4 is outside the tolerance of 1 from an expected value of 2 🚨``` 
or
```
>       assert len(run_nanorc.data_files) == expected_number_of_data_files
E       AssertionError: assert 3 == 4
```

Which probably means more checks with RTCM are needed to confirm it definitely works... I don't understand how we get 4 events (maybe there's another bug hiding in RTCM somewhere?) or 3 data files where there's clearly 4 tests.

bieryAtFnal (approved):
LGTM

PR #359, opened 2024-11-05
Add a test of the change-rate command to RandomTriggerCandidateMaker

eflumerf:


bieryAtFnal (approved):
LGTM.
There are errors reported when running this test on the develop branch, as expected, since the develop branch does not have Artur's most recent RTCM fixes.

Issue #278, opened 2024-02-29
OKS: get clock frequency from Session's detector_configuration


ArturSztuc:
_Originally posted by @gcrone in https://github.com/DUNE-DAQ/trigger/pull/277#discussion_r1507934805_

ArturSztuc:
Similar issue for the readoutlibs (@glehmannmiotto @gcrone, https://github.com/DUNE-DAQ/readoutlibs/pull/152#discussion_r1507532071), with an idea to retrieve detector config from ModuleConfiguration. Searching through DUNEDAQ, I only found mention of `ModuleConfiguration` [HERE](https://github.com/DUNE-DAQ/docs/blob/b80a3ac281b90c6a80560d736bbdee443fcf9b2e/docs/packages/appfwk/appfwk.drawio#L79):

gcrone:
ModuleConfiguration is [here](https://github.com/DUNE-DAQ/appfwk/blob/develop/include/appfwk/ModuleConfiguration.hpp) and [here](https://github.com/DUNE-DAQ/appfwk/blob/develop/src/ModuleConfiguration.cpp)

ArturSztuc:
1. `RandomTCMaker` overhaul (merged): https://github.com/DUNE-DAQ/trigger/pull/350
2. `CustomTCMaker: removal (PR pending): https://github.com/DUNE-DAQ/trigger/pull/364

Once the second PR is merged, this issue can be closed.

PR #364, opened 2024-11-14
Removing CustomTCMaker,  replaced by vector of RandomTCMakers

ArturSztuc:
Removing `CustomTCMaker` from the trigger.

This PR goes together with:
1. https://github.com/DUNE-DAQ/appmodel/pull/156
2. https://github.com/DUNE-DAQ/daqsystemtest/pull/161

`CustomTCMaker` was useful for testing e.g. MLT logic: it was capable of outputting. It was capable of outputting any number of TC-types, at different time-intervals from each other, with different time windows.

`RandomTCMaker` now has fully-customizable time window and TC-output (as well as trigger rate). With v5, we can have multiple `RandomTCMaker`s running concurrently, each with different configuration, making `CustomTCMaker` obsolete.

Tested by running multiple `RandomTCMakers` concurrently with different redout windows/rates/types, and looking at the output HDF5 file, comparing against similarly-configured `CustomTCMaker` runs.


ArturSztuc:
Together with https://github.com/DUNE-DAQ/trigger/pull/350, resolves issue: https://github.com/DUNE-DAQ/trigger/issues/278

aeoranday (approved):
LGTM. Tested online and with emulation.

MRiganSUSX:
Two issues were discovered with more testing of running with multiple RTCM (or generally multiple StandaloneTCMaker [STCM]) instances.

- When > 1 STCM are set to listen to `kTimeSync` messages, issue described in https://github.com/DUNE-DAQ/iomanager/issues/70 is observed. This should be resolved with https://github.com/DUNE-DAQ/iomanager/pull/79.

- When > 2 STCM makers are used - both using `kSystemTime` for times - there is a seg fault error from `mlt` at `stop-trigger-sources`. This should be fixed with https://github.com/DUNE-DAQ/trigger/pull/367.

More thorough testing to happen after the fixes above are in place.

PR #366, opened 2024-11-14
Updated uses of the string_to_fragment_type_value() method to use its…

bieryAtFnal:
… new name, string_to_trigger_candidate_type(), which seems much more appropriate.

These changes need to be tested and merged along with corresponding changes in the [trgdataformats](https://github.com/DUNE-DAQ/trgdataformats/pull/47) and [triggeralgs](https://github.com/DUNE-DAQ/triggeralgs/pull/88) repos.


ArturSztuc (approved):
LGTM

Issue #354, opened 2024-10-30
Silent timesync errors when configuration is inconsistent 


roland-sipos:
- The trigger IF not receiving timesyncs (despite configured to do so in the RandomTCMaker, will silently log gazzilions of ERRORs (not visible on opmon) and send absurd windows.
   + Proper error handling is needed here. The trigger should send these errors to Kafka.

- The RandomTC maker can also switch to system clock based TS-ing, instead of timesync. But when we attempted to STOP, the trigger app crashed. (What I saw is that some threads were not properly joined/stopped, but destroyed. )
   + We should check if the stop scenario is fine with either timesync and sysclock based decisions.

We encountered these errors when config issues and incosistencies were present on EHN1 DAQ configs.
To be cross-checked if the stop logic is intact with sysclock based estimation and failed timesync reception should send errors to Kafka/Opmon

ArturSztuc:
The first issue is linked to this one in iomanager: https://github.com/DUNE-DAQ/iomanager/issues/48

ArturSztuc:
The second issue is resolved via this PR: https://github.com/DUNE-DAQ/trigger/pull/367

ArturSztuc:
Assigning to v5.3.0, since issue https://github.com/DUNE-DAQ/iomanager/issues/48 is assigned to that release too.

ArturSztuc:
The first issue now also seems resolved in
https://github.com/DUNE-DAQ/iomanager/issues/48 and https://github.com/DUNE-DAQ/iomanager/pull/113

PR #367, opened 2024-11-14
Fixing RTCM shutdown & removing few obsolete pars

ArturSztuc:
Removed few obsolete `RandomTCMaker` members & fixed bug in the shutdown when running with `kSystemClock` instead of `kTimeSync`. 

Tested by running test session with `kTimeSync`, receiving correct number of TCs and shuts down without segfault.
Fixes second part of issue: https://github.com/DUNE-DAQ/trigger/issues/354


aeoranday (approved):
LGTM. Tested online and by emulation.

PR #368, opened 2024-11-19
Updating RTCM receiver pubsub to use IOM receiver tags

ArturSztuc:
Requires the following PRs:
1. https://github.com/DUNE-DAQ/iomanager/pull/79
2. https://github.com/DUNE-DAQ/listrev/pull/63

Fixes issue in a comment described in https://github.com/DUNE-DAQ/trigger/pull/364, which is a result of https://github.com/DUNE-DAQ/iomanager/issues/70.

In short: Only one subscriber was allowed in pubsub connection. When trying to run with two `RandomTCMaker`s, both using `kTimeSync`, they would both try to connect response/callback function to the same `TimeSync` connection pubsub, resulting in a crash.

The PRs above add new feature in `IOManager` to allow multiple receivers for a connection, and this PR makes `RandomTCMaker`s using this new feature.

Tested by running two `RandomTCMakers` with `kTimeSync`, and checkout the log messages / output HDF5 file.
Might need few other packages (e.g. `appfwk`) for others to test.

### DIFF
--- a/plugins/RandomTCMakerModule.cpp
+++ b/plugins/RandomTCMakerModule.cpp
@@ -62,7 +62,7 @@ RandomTCMakerModule::init(std::shared_ptr<appfwk::ModuleConfiguration> mcfg)
   // Get the input connections
   for(auto con: mtrg->get_inputs()) {
     // Get the time sync source
-    TLOG() << "Timestamp receiver connection is " << con->class_name() << "@"
+    TLOG() << "TimeSync receiver connection is " << con->class_name() << "@"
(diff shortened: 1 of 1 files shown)


eflumerf (approved):
Builds and the regression tests still pass. Only concern is since this depends on `iomanager`, it will be tricky to incorporate into existing areas until the nightly is cut.

ArturSztuc:
Tested again with the new nightly. Doesn't require `iomanager` (and all of its dependencies) to be pulled through anymore, so won't cause build issues for people pulling the `trigger` repository.
Merging.

PR #370, opened 2024-11-20
Updated the change_rate_test to use the new drunc --run-number and --…

bieryAtFnal:
…trigger-rate command options.

Without these changes, the change_rate_test does not work.
With these changes, hopefully it does work.  (It did for me.)


eflumerf (approved):
Test works with change, doesn't without

PR #372, opened 2024-12-16
TPRequestHandler: check whether sink queue is set

MRiganSUSX:
Added small check that a sink queue is set before proceeding with periodic data transmission.


glehmannmiotto (approved):
LGTM

PR #380, opened 2025-02-26
Add "tc_time_outside_window_test", which is just

eflumerf:
small_footprint_quick_test from daqsystemtest with readout window specified in trigger such that the td time is outside the window for the TC fragment.

(See https://github.com/DUNE-DAQ/daqsystemtest/pull/183 for update to small_footprint_quick_test now that the issue demonstration has been moved here)


bieryAtFnal (approved):
LGTM.
(As discussed with Eric, I renamed the new test from td_* to tc_*, since it is the TC that is outside the request window.)

ArturSztuc (approved):
LGTM, test working as intended

PR #382, opened 2025-03-06
TP Replay Application (v5)

MRiganSUSX:
This needs these other PRs:
[appmodel](https://github.com/DUNE-DAQ/appmodel/pull/173)
[daqsystemtest](https://github.com/DUNE-DAQ/daqsystemtest/pull/186)

Details here: https://github.com/DUNE-DAQ/trigger/tree/mrigan/trigger_replay_script/python/trigger/tpreplay_application


MRiganSUSX:
I have merged develop (with 5.3 changes) in.

I'm now considering the changes that will come with PDS TPs...

MRiganSUSX:
Now updated to work with new TP format.
After discussion with Wes it was suggested to currently limit replay to TPC TPs only until PDS TPs are fully integrated in trigger (so that we can reflect that exactly in replay), and that was done in the last few commits.
Ready to review (again).

CMakeLists.txt
```diff
@@ -52,9 +52,9 @@ daq_add_library(
 ## As a one liner, this call is a major source of git conflicts due to
 ## us all changing it.  Please keep one file per line to give git
 ## merge a chance to succeed.
-daq_codegen(
-  triggerprimitivemaker.jsonnet
-  TEMPLATES Structs.hpp.j2 Nljs.hpp.j2 )
+#daq_codegen(
+#  triggerprimitivemaker.jsonnet
+#  TEMPLATES Structs.hpp.j2 Nljs.hpp.j2 )
```
ArturSztuc: Should we just remove this?

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+  auto mtrg = mcfg->get_dal<appmodel::TPReplayModule>(get_name());
+  m_conf = mtrg->get_configuration();
+  if (!m_conf) {
+    throw ReplayConfigurationProblem(ERS_HERE, get_name(), "Missing configuration!");
+  }
+
+  // ### Extract relevant objects
+  // Clock speed
+  clocks_per_us = mcfg->session()->get_detector_configuration()->get_clock_speed_hz() /
+                  1'000'000.0; // please keep the '.' to make this float
```
ArturSztuc: Do you mean a double? If you want this to be a float then you might need:
`1'000'000.0F` (note the `F` that makes this a float). Could also be more explicit with `float(1'000'000)` or some other casting.

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+  TLOG() << "### REPLAY CONFIGURATION ###";
+  TLOG() << "Will use channel map: " << m_channel_map_name;
+  try {
+    m_channel_map = dunedaq::detchannelmaps::make_map(m_channel_map_name);
+  } catch (const detchannelmaps::ChannelMapCreationFailed& e) {
+    ers::error(dunedaq::trigger::ReplayChannelMapProblem(ERS_HERE, get_name(), m_channel_map_name));
+  }
+
+  // Loops
+  m_loops = m_conf->get_number_of_loops();
```
ArturSztuc, line 90: Perhaps it would be good to have this as an option, I can imagine users wanting to just run drunc for some time (few minutes, up to few hours even) to see the stability of their algorithms/changes.

Could e.g. have `m_loops` to be `-1` as default, which would be a condition to "just keep on running till user stops drunc"

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+
+  // Plane filtering
+  for (auto& plane_wrap : m_conf->get_filter_out_plane()) {
+    m_filter_planes_ids.push_back(plane_wrap->get_plane());
+  }
```
ArturSztuc: Could just be a vector of `int`s instead

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+  auto con = mtrg->get_outputs();
+
+  // Global times
+  m_earliest_first_tp_timestamp = std::numeric_limits<triggeralgs::timestamp_t>::max();
+  m_latest_last_tp_timestamp = 0;
+
+  // Loading sorted TP stream files
+  for (auto& stream : m_conf->get_tp_streams()) {
+    std::pair tmp_obj = std::make_pair(stream->get_index(), stream->get_filename());
+    m_tpstream_files.insert(tmp_obj);
```
ArturSztuc: Since it's a map, shouldn't it be possible to just do something like:
`m_tpstream_files[stream->get_index()] = stream->get_filename();`, rather than creating intermediate pair object?

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+    ers::error(dunedaq::trigger::ReplayNoValidFiles(ERS_HERE, get_name()));
+  }
+
+  // Data loaded and sorted.
+  // Now we create streams.
+  int global_iter = 0;
+  // Loop over ROUs
+  for (const auto& rou_pair : m_all_tp_data) {
+    const std::string& ROU = rou_pair.first;
+    const auto& plane_map = rou_pair.second;
```
ArturSztuc:
```suggestion
  for (const auto& [ROU, plane_map] : m_all_tp_data) {
```

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+    int plane_iter = 0;
+    // Loop over Planes
+    for (const auto& plane_pair : plane_map) {
+      int plane = plane_pair.first;
+      const auto& vector_of_tps = plane_pair.second;
```
ArturSztuc:
```suggestion
    for (const auto& [plane, vector_of_tps] : plane_map) {
```

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+
+      m_earliest_first_tp_timestamp =
+        std::min(m_earliest_first_tp_timestamp, this_stream.tpvs.front().front().tp.time_start);
+
+      m_latest_last_tp_timestamp = std::max(m_latest_last_tp_timestamp, this_stream.tpvs.back().back().tp.time_start);
+
+      m_tp_streams.push_back(std::move(this_stream));
+      plane_iter++;
+    }
+    global_iter = global_iter + plane_iter;
```
ArturSztuc: Why not the standard
```suggestion
    global_iter += plane_iter;
```

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+    for (const auto& plane_pair : plane_map) {
+      int plane = plane_pair.first;
+      const auto& vector_of_tps = plane_pair.second;
+
+      TPStream this_stream;
+      TLOG_DEBUG(1) << "Stream: " << (global_iter + plane_iter) << "; ROU: " << ROU << "; plane: " << plane
+                    << "; TP sink is " << con[global_iter + plane_iter]->class_name() << "@"
+                    << con[global_iter + plane_iter]->UID();
+      this_stream.tp_sink =
+        get_iom_sender<std::vector<trigger::TriggerPrimitiveTypeAdapter>>(con[global_iter + plane_iter]->UID());
```
ArturSztuc: What happens when we have e.g. 8 files from NP04, 2 consecutive ones for each APA?
It looks like it will generate 8*3 TPAdapters, rather than 4*3 (or instead of crashing gracefully, if we cba to implement it at this point)

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+  // Reset opmon
+  m_tp_made_count.store(0);
+  m_tpv_made_count.store(0);
+  m_tpv_failed_sent_count.store(0);
+
+  // We need the wall-clock time at which we'll send out the TPs
+  // with the earliest timestamp, so we can keep all of the output
+  // streams in sync. We pick "now" plus a bit, to allow time for all
+  // of the threads to start up
+  auto earliest_timestamp_time = std::chrono::steady_clock::now() + std::chrono::milliseconds(100);
```
ArturSztuc, line 187: Nice

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+      } catch (...) {
+        ers::error(dunedaq::trigger::ReplayPlaneError(ERS_HERE, get_name(), filename));
+        continue;
+      }
+
+      // hack for APA1, basically making plane 1 collection plane
+      // decide whether we want this here long term
+      if (ROU == "APA_P02SU") {
+        plane = (plane == 1) ? 2 : (plane == 2) ? 1 : plane;
+      }
```
ArturSztuc, line 364: I think it's fine as long as it's in README.md, although the readme is so long I doubt many people will read it.
Could be worth to either issue a warning here, or maybe just remove and let people sort this out in the .data.xml themselves.

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+        continue;
+      }
+
+      // Extract trigger primitives
+      // Create a vector of the correct size, and directly associate it with tp_array
+      std::vector<TriggerPrimitiveTypeAdapter> tps(reinterpret_cast<TriggerPrimitiveTypeAdapter*>(tp_array),
+                                                   reinterpret_cast<TriggerPrimitiveTypeAdapter*>(tp_array) + num_tps);
+
+      // Efficient insertion into deque
+      auto& data_deque = all_data[ROU][plane];
```
ArturSztuc, line 381: Deque is cheap to write, and expensive to read.
Nothing to change here, just a comment. Might be worth checking if this slows down the processing of the TPs down the line significantly, or effectively does not matter.

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+        data_deque.push_back(std::move(tps));
+      }
+      // Append if already in order
+      else if (data_deque.back().front().tp.time_start <= tps.front().tp.time_start) {
+        data_deque.push_back(std::move(tps));
+      }
+      // Prepend if this vector is older than the first element
+      else if (data_deque.front().front().tp.time_start >= tps.front().tp.time_start) {
+        data_deque.push_front(std::move(tps));
+      }
```
ArturSztuc, line 394: The `back().front()` is pretty confusing. So the end of the very first vector?
I think this should either be .back().back() with `front().front()` or `.back().front()` and `front().back()`, but I'm not sure... I would like to test this, but the code does not compile or run (see general comment) :|

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+
+  } // files loop
+
+  TLOG() << "Data loading summary (all):";
+  TLOG() << "------------------------------";
+  TLOG() << "Files: " << m_tpstream_files.size();
+  // Loop through the map and print sizes
+  for (const auto& rou_pair : all_data) {
+    const std::string& ROU = rou_pair.first; // ROU name (key)
+    const auto& plane_map = rou_pair.second; // Map of planes for this ROU
```
ArturSztuc:
```suggestion
  for (const auto& [ROU, plane_map] : all_data) {
```

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+
+    TLOG() << "ROU: " << ROU << ", Number of planes: " << plane_map.size();
+
+    // Loop through each plane for the current ROU
+    for (const auto& plane_pair : plane_map) {
+      int plane = plane_pair.first;                  // Plane number (key)
+      const auto& vector_of_tps = plane_pair.second; // Vector of TriggerPrimitiveTypeAdapter vectors
```
ArturSztuc:
```suggestion
    for (const auto& [plane, vector_of_tps] : plane_map) {
```

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+  int local_tp_made = 0;
+  int local_tpv_made = 0;
+  int local_tpv_failed = 0;
+
+  while (running_flag.load()) {
+
+    // Looping logic
+    if (current_iteration >= m_loops) {
+      break;
+    }
```
ArturSztuc: Would be nice to have "run till user says stop" logic

plugins/TPReplayModule.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+namespace dunedaq {
+namespace trigger {
+class TPReplayModule : public dunedaq::appfwk::DAQModule
+{
+public:
+  /**
+   * @brief RandomDataListGenerator Constructor
+   * @param name Instance name for this RandomDataListGenerator instance
+   */
```
ArturSztuc: I don't think this comment matches the constructor function...

We also should be using doxygen comments everywhere nowadays.

plugins/TPReplayModule.cpp
```diff
@@ -0,0 +1,551 @@
...
+  m_run_start_time = std::chrono::steady_clock::now();
+
+  // Start threads for each stream
+  for (auto& stream : m_tp_streams) {
+    m_threads.push_back(std::make_unique<std::thread>(&TPReplayModule::do_work,
+                                                      this,
+                                                      std::ref(m_running_flag),
+                                                      std::ref(stream.tpvs),
+                                                      std::ref(stream.tp_sink),
+                                                      earliest_timestamp_time));
```
ArturSztuc, line 197: I imagine the amount of time taken to create these could depend on the size of `stream.tpvs`, and so with large large / many TPStream files maybe we could take more than 100ms in total. Was it tested? I cannot test it now (general comment)

python/trigger/tpreplay_application/README.md
```diff
@@ -0,0 +1,373 @@
...
+- *Init stage*: A lot is happening inside the `TPReplayModule` at the init stage: parsing configuration, multiple checks on HDF5 files, extracting ROUs, actually extracting TP data, plane filtering... Depending on the number of files this can take a lot of time. If needed, portions of this can be moved to different run stages.
+- *Expectations ?*: There are many places in the current dune-daq code where expectations are baked in (but not necessarily documented), for example, an expectation for queues that are at times not obvious (ie `TPRequestHandler` is expected to link to `FragmentAggregatorModule`, but this module is not required outside readout).
+- *Memory limits*: Some memory optimization is implemented, however, TPStream files are often very big. Because most processing happens within one module (TPMm), memory usage can be an issue for many files at once. You have been warned.
+
+### TODO (future)
+- [ ] should the initial TP times be shifted (as if they were streamed now) ?
+- [ ] option in python script to pick TA algorithm ?
+- [ ] support for multiple concurrent (different) makers
+- [ ] when TP format changes (relative `tp.time_peak`) looping logic needs adjusting
```
ArturSztuc: The format has already changed I believe. Does this still need changing?

python/trigger/tpreplay_application/__main__.py
```diff
@@ -0,0 +1,381 @@
...
+
+def set_mem_limit(verbose: bool):
+    """
+    As a safety measure we set a memory limit for the process.
+    Should not be needed here as hdf5 processing is minimal.
+    """
+    GB = 1024**3
+    memory_limit = 25 * GB 
+    resource.setrlimit(resource.RLIMIT_AS, (memory_limit, memory_limit))
+    logging.debug("Setting memory limit to %i GBs", memory_limit/GB)
```
ArturSztuc: Probably should be adjustable, there are machines with more than 25gigs

python/trigger/tpreplay_application/__main__.py
```diff
@@ -0,0 +1,381 @@
...
+    with open(filename, "r") as file:
+        tpstream_files = [line.strip() for line in file.readlines()]
+    cleanup()
+
+    logging.info("Total files to process: %i", len(tpstream_files))
+    logging.debug("TPStream files loaded: %s", tpstream_files)
+
+    return tpstream_files
+
+def check_files(files):
```
ArturSztuc: Is it `files: List[str]`?

python/trigger/tpreplay_application/__main__.py
```diff
@@ -0,0 +1,381 @@
...
+        path = Path(a_file)
+        if not path.exists():
+            logging.error("File %s does not exist!", a_file)
+            sys.exit(1)
+        # check it's hdf5
+        if not path.suffix.lower() in ('.h5', '.hdf5'):
+            logging.error("File %s does not seem to be hdf5 file!", a_file)
+            sys.exit(1)
+
+def extract_rous_and_planes(files: List[str], channel_map, planes_to_filter: Set[int], verbose: bool) -> (List[TPStreamFile], ROUPlaneData):
```
ArturSztuc: Strange that hannel map the only one without type hints

python/trigger/tpreplay_application/__main__.py
ArturSztuc, line 1: Not important, but the type hints are not consistent throughout: some function arguments have them, some don't.

Type hints like `List`, `Set`, `Dict` etc are now deprecated, as [commented](https://docs.python.org/3/library/typing.html#typing.Dict) by Alex in this PR: https://github.com/DUNE-DAQ/trgtools/pull/56#discussion_r2000532849

Many functions seem to have `verbose` passed as an argument, but then do not use it. Is this by desgin?

python/trigger/tpreplay_application/__main__.py
```diff
@@ -0,0 +1,381 @@
...
+import copy
+import gc
+import logging
+import re
+import resource
+import sys
+from collections import defaultdict
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Dict, Set, List
```
ArturSztuc: Deprecated, see comment attached to this file

python/trigger/tpreplay_application/__main__.py
```diff
@@ -0,0 +1,381 @@
...
+
+            if i == 0:
+                all_tpstream_files.append(TPStreamFile(tpstream_file, tp.time_start, 0))
+                logging.debug("First time start: %s", tp.time_start)
+
+            # check subdetector
+            subdet = tp.detid
+            print(subdet)
+            if subdet not in valid_subdetectors:
+                continue
```
ArturSztuc:
```suggestion
            if tp.detid not in valid_subdetectors:
                continue
```

I think it's that print line that caused many 0s being printed out when I tried running this code. So either remove, or maybe add extra info to the `print()` function so people know what these numbes are :)

python/trigger/tpreplay_application/__main__.py
```diff
@@ -0,0 +1,381 @@
...
+    As a safety measure we set a memory limit for the process.
+    Should not be needed here as hdf5 processing is minimal.
+    """
+    GB = 1024**3
+    memory_limit = 25 * GB 
+    resource.setrlimit(resource.RLIMIT_AS, (memory_limit, memory_limit))
+    logging.debug("Setting memory limit to %i GBs", memory_limit/GB)
+
+def cleanup():
+    gc.collect()
```
ArturSztuc: This might make the future development difficult and full of mines, but removing it now means re-structuring everything... so let's keep this for now, but remember to figure out how to rewrite without it.

python/trigger/tpreplay_application/__main__.py
```diff
@@ -0,0 +1,381 @@
...
+    parser.add_argument("--config", type=str, default="config/daqsystemtest/example-configs.data.xml", help="Path to OKS configuration file.")
+    parser.add_argument("--path", type=str, default="tpreplay-run", help="Path for local output for configuration files.")
+    parser.add_argument("--verbose", action="store_true", help="Enable verbose logging.")
+    args = parser.parse_args()
+
+    # Set up logging based on the verbose flag
+    setup_logging(args.verbose)
+
+    # Set memory limit
+    set_mem_limit(args.verbose)
```
ArturSztuc: Could make this optional, but enabled by default?

ArturSztuc (changes requested):
Mostly minor comments, and one large one:

I did not manage to finish the review. The code currently does not compile on recent release. I tried doing git merge: after resolving some conflicts, I managed to run the python script, but not drunc with the configuration generated, it was crashing at boot. 
I will continue the PR when this works, as I would like to test few pieces of logic in the code live, and, of course, would like to make sure it works :)

plugins/TPReplayModule.cpp
MRiganSUSX: `m_filter_planes_ids` is a vector of `int`s ?
```std::vector<int> m_filter_planes_ids``` in header.

Unless you mean setting it directly without the loop: I have tried that and it did not work without additional casting. Given we will only have at most 2 planes to filter, the loop is not very limiting here.

File: plugins/TPReplayModule.cpp (hunk 5 above)
Comment (MRiganSUSX) [reply plugins/TPReplayModule.cpp]: This would be easier to read but would result in overwrites if the index is accidentally the same. I have extended the logic to catch that and throw an issue otherwise.

File: plugins/TPReplayModule.cpp (hunk 11 above)
Comment (MRiganSUSX) [reply plugins/TPReplayModule.cpp:364]: This is mentioned in the README, but could perhaps be highlighted more. 

I'm not sure what you mean by setting this in configuration. Unfortunately, the filtering is currently not per APA but global - would you want to add that to the list of TODOs?

File: plugins/TPReplayModule.cpp (hunk 13 above)
Comment (MRiganSUSX) [reply plugins/TPReplayModule.cpp:394]: (remember the structure is `std::deque<std::vector<TriggerPrimitiveTypeAdapter>>`)

I think the logic is correct. Basically I tried to optimize the inserts using cases, where the ordering is always given by the tp.time_start of the FIRST TP in the vector (each vector is a fragment so the rest of TPs in this vector are ordered):
- if the deque is empty -> just insert
- if the data that is to be inserted is newer [newer = bigger timestamp] (by comparing to the last in the deque - which is `.back()`) -> append
- if the data that is to be inserted is older [older = smaller timestamp] (by comparing to the first in the deque - which is `.front()`) -> prepend 
- in other case (not newer nor the oldest; which should be rare): do a binary search.

I have written a loop to check the ordering after the data is loaded (just a test, not adding it to the code) and it confirmed it's ordered. I have also tried to simply print the timestamps and just sanity check and I think that was also correct but it was rough (too many objects).

check loop if you'd like to confirm:
```
   bool is_ordered = true; // Flag to track if the data is ordered
 
   // Loop over the map
   for (const auto& outer_pair : m_all_tp_data) {
       for (const auto& inner_pair : outer_pair.second) {
           const auto& data_deque = inner_pair.second; // deque<std::vector<TriggerPrimitiveTypeAdapter>>
 
           // Loop over the deque and check if time_start is ordered
           for (size_t i = 1; i < data_deque.size(); ++i) {
               const auto& prev_tp = data_deque[i - 1].front().tp;
               const auto& curr_tp = data_deque[i].front().tp;
 
               if (prev_tp.time_start > curr_tp.time_start) {
                   std::cerr << "Order violation detected: "
                             << "prev_time_start = " << prev_tp.time_start
                             << ", current_time_start = " << curr_tp.time_start << std::endl;
                   is_ordered = false; // Set flag to false if an order violation is found
                   break; // Exit the loop early on violation
               }
           }
 
           if (!is_ordered) break; // Exit outer loop if order is violated
       }
 
       if (!is_ordered) break; // Exit the outermost loop if any violation is found
   }

  // Final check
  if (is_ordered) {
    std::cout << "All data is ordered correctly!" << std::endl;
  } else {
    std::cout << "Some data is out of order!" << std::endl;
  }
```

but please double check this as it's *quite* important.

File: plugins/TPReplayModule.cpp (hunk 18 above)
Comment (MRiganSUSX) [reply plugins/TPReplayModule.cpp:197]: This is a good point. I have not seen issues with this so far, but wasn't specifically looking for this.

It is not realistically possible to have too large / too many TPSream files as they use a lot of memory, so there is a global system limit. 

No issue to extend 'timeout' if needed.

File: python/trigger/tpreplay_application/__main__.py (hunk 26 above)
Comment (MRiganSUSX) [reply python/trigger/tpreplay_application/__main__.py]: I don't think these should be optional? 
- the logging should always be there (as feedback), the user can choose verbose ON/OFF. When verbose is OFF there are very few 'summary' lines
- the mem limit is now configurable (after your comment earlier)

do you think there should be an option to completely turn any one or both of these off ?

File: plugins/TPReplayModule.cpp (hunk 9 above)
Comment (MRiganSUSX) [reply plugins/TPReplayModule.cpp]: The logic of having one tpstream file per readout unit is no longer being adhered to ( :disappointed: ). 

Regardless, I understand your point. When the data is looped over and the ROUs and planes extracted, they are stored in sets; and only *unique* sourceIDs are used. So if we want to replay data from a run, and we have say 3 files, each containing all ROUs (it doesn't matter actually whether it is 1 or all, the looping checks every fragment now), we will end up with a map of [ROU: plane: data]. Each ROU:plane combination will be a separate/unique source of data. So 3 consecutive files, each containing 2ROUS and 3PLANES, we end up with 6 unique source = 6 streams. This number of connections is created by the python script and the cpp code just picks one incrementally from a vector of these.

**Comment by MRiganSUSX** [comment]
Thanks a lot for the detailed review @ArturSztuc.
Apologies for this not building - I've tried to keep it up to date with develop but it was an active development period.

Regarding your comments:
- Anything that is marked as resolved was addressed and is part of the new commits.
- For a few of the comments I responded with more details, please have a look at responses. I'm happy to do more changes but perhaps we should discuss some of these a little more. 

Brief summary:
- dev merged, fix to build
- removed obsolete parts of cmakelist
- removed obsolete python includes
- clock value casted for safety
- improved for loops
- updated comments
- memory limit now a configurable parameter
- type hints added throughout
- cleanup of the verbose functionality
- option to set n-loops, -1 by default: Logic propagated to TPReplay module to 'keep replaying'.
- updated README with these changes
- other, not requested features:
-- RTCM rate set to 0
-- improved logging

Otherwise, I ran several tests including filtering, different channel maps, various n-loops values, and different input data (np02, np04, consecutive hdf5s, ...). I did not encounter any new issues, so ready for another review.

---
One more thing I am playing with right now is shifting the TP times to be dynamic 'from the first TP'. This will be on a separate branch if it turns out to be useful.

**Comment by MRiganSUSX** [comment]
If you have time, there is also a branch https://github.com/DUNE-DAQ/trigger/tree/mrigan/replay_shift that is based on this but adds shifting of TP times initially so that they appear 'now'.

File: plugins/TPReplayModule.cpp (hunk 18 above)
Comment (ArturSztuc) [reply plugins/TPReplayModule.cpp:197]: Not for this PR, but for the future one, once we have far more data (e.g. from np02) we need to remember the 100ms rule and see if it needs to change to a configurable, or maybe even based off the number of files being processed. Future PR tho, not this :)

File: plugins/TPReplayModule.cpp [hunk 27]
```diff
@@ -53,7 +53,7 @@ TPReplayModule::init(std::shared_ptr<appfwk::ConfigurationManager> mcfg)
   // ### Extract relevant objects
   // Clock speed
   clocks_per_us = mcfg->session()->get_detector_configuration()->get_clock_speed_hz() /
-                  1'000'000.0; // please keep the '.' to make this float
+                  double(1'000'000.0); // this is redundant but safer...
```
ArturSztuc: Ah, so you wanted a double! Sorry, I was confused by the earlier `float`.

plugins/TPReplayModule.cpp
```diff
@@ -81,9 +79,7 @@ TPReplayModule::init(std::shared_ptr<appfwk::ConfigurationManager> mcfg)
   m_loops = m_conf->get_number_of_loops();
 
   // Plane filtering
-  for (auto& plane_wrap : m_conf->get_filter_out_plane()) {
-    m_filter_planes_ids.push_back(plane_wrap->get_plane());
-  }
+  m_filter_planes_ids = std::set<int>(m_conf->get_filter_out_plane().begin(), m_conf->get_filter_out_plane().end());
```
ArturSztuc, line 93: Vector didn't work? I thought `m_conf->get_filter_out_plane()` would already output `std::vector`.

ArturSztuc (approved):
Thanks a lot for this, LGTM!

The main comment is in https://github.com/DUNE-DAQ/appmodel/pull/173.

Let's merge this :)

MRiganSUSX:
Hi @ArturSztuc, 
Thanks again for the review and comments.

I have gone through another little refactor: due to external changes, and due to the issue you discovered with the example configurations and segments allocation. 

#### Summary of changes:
- trigger:
  - updated replay readme regarding daqsystemtest changes (segments, sessions)
  - updated general trigger repo readme with a link to replay
  - since we merged in the conf/scrap PR, I also propagated this to replay (no functional changes)
- appmodel:
  - Alessandro did a refactor there regarding object factory, so I had to propagate that for replay...
- daqsystemtest:
  - I managed to solve the allocation issue (thanks for spotting that) by creating a tpreplay-segment (that only contains TPReplayApp) and tpreplay-root-segment which contains the minimal set-up needed for replay (replay, trigger, df apps). This way we can offer example replay sessions without affecting the rest of example sessions.

MRiganSUSX:
As as side-note, I have opened issues for all the comments/suggestions you had that were not to be part of this PR:
- https://github.com/DUNE-DAQ/trigger/issues/392
- https://github.com/DUNE-DAQ/trigger/issues/393
- https://github.com/DUNE-DAQ/trigger/issues/394

ArturSztuc (approved):
LGTM! 

Sorry it took a while. I initially couldn't get it to work because I had `patch/fddaq-v5.3.x` locally... works under fresh `development` though.

Issue #381, opened 2025-03-05
Obsolete python apps


MRiganSUSX:
While updating TP replay, I've noticed the use of PrimitiveMaker across multiple python apps. Looking at those apps more closely, they are obsolete and non-functional in v5 as they use modules and plugins that are no longer used/present in trigger (like zipper, fake heartbeats...).

These should be removed or reworked to function in v5.

Apps:
```
/trigger/python/trigger/dbscan_ta_to_sink/
/trigger/python/trigger/faketp_to_sink/
/trigger/python/trigger/test_channel_filter/
```

and also external to trigger: 
```/dfmodules/python/dfmodules/tpset_demo/```
this seems to make use of TriggerPrimitiveMaker that relied on .txt files (a while back)...

ArturSztuc:
@MRiganSUSX Could you remove these? I think we could do this before v5.3.0

MRiganSUSX:
Now addressed in PR https://github.com/DUNE-DAQ/trigger/pull/385

MRiganSUSX:
Fixed by the PR mentioned above.

PR #385, opened 2025-03-25
Removing obsolete apps

MRiganSUSX:
This addresses issue https://github.com/DUNE-DAQ/trigger/issues/381

There were several python applications left over from v4 that no longer work with v5 for several reasons:
- using now nonexistent jsonnet files
- using modules that no longer exist (zipper, heartbeatmaker, channelfiler, ...)
- using modules that were reworked and they expect the old version (TriggerPrimitiveMaker, apps using txt version from years ago)
- using v4 daqconf functions
- generally not using OKS (for places where it is in use now)
...

Associated PR: https://github.com/DUNE-DAQ/dfmodules/pull/405


ArturSztuc (approved):
LGTM, thanks for cleaning this up!

PR #387, opened 2025-03-26
Implement TP serialization using custom serialization routines.

eflumerf:


aeoranday (approved):
LGTM. Tested on both unit tests and integ. tests.

eflumerf:
Adding some additional detail...this PR was done so that `iomanager` and `serialization` do not have to change their APIs. Implementing a custom JSON serializer for the TriggerPrimitive class gets around the issue where the default serializer cannot handle bit-fields.

PR #390, opened 2025-05-08
Adding scrap for TXProcessor

MRiganSUSX:
An issue was reported where the triggeralgs processors were not reaching their destructors. 
The issue is that the TXProcessors are not set up to clear their vectors (often of pointers). This is especially problematic for sequences of multiple runs in the same session where terminate is not reached at run change (ie -> stop -> scrap -> conf -> start (new run)). 

There are no logic changes, but the scrap method is made available for each TXProcessor.

Tested with local sessions and suit of integration tests. 

Can be tested with https://github.com/DUNE-DAQ/triggeralgs/compare/develop...mrigan/ta_makers_debug (that is not to be merged, just for testing).

### DIFF
--- a/src/TAProcessor.cpp
+++ b/src/TAProcessor.cpp
@@ -102,6 +102,14 @@ TAProcessor::conf(const appmodel::DataHandlerModule* conf)
 }
 
 void
+TAProcessor::scrap(const nlohmann::json& args)
+{
+  inherited::scrap(args);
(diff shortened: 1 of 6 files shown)


MRiganSUSX:
I decided to make a bigger overhaul of the conf/scrap methods for all modules in trigger, these changes will be absorbed there...

PR #391, opened 2025-05-12
Proper conf and scrap implementation for trigger modules

MRiganSUSX:
There was an issue discovered where destructors of makers (triggeralgs modules) were not being reached. There were multiple reasons for this (the base classes in datahandlinglibs not reaching destructors ?) but another one is that the scrap methods were not (fully) implemented in some trigger modules.
Another reason this is important is the potential `stop -> scrap -> conf` fsm transition which is expected to work (currently doesn't due to issues in df). 

Here is the implementation of more appropriate conf and scrap methods for the trigger modules.
There are no logical / functional changes, the code changes are just splitting already present functionality to relevant init,conf,scrap blocks; additional cleanup in the scrap blocks, and some improved logging. 

Tested locally (one can use https://github.com/DUNE-DAQ/triggeralgs/compare/develop...mrigan/ta_makers_debug for testing only) and passed the usual suite of integration tests.

### DIFF
--- a/plugins/MLTModule.cpp
+++ b/plugins/MLTModule.cpp
@@ -106,7 +106,7 @@ MLTModule::do_configure(const nlohmann::json& /*obj*/)
       SubdetectorID detid = static_cast<SubdetectorID>(gidmap["detector_id"]);
       if (m_srcid_detid_map.contains(sourceid) && !(m_srcid_detid_map[sourceid] == detid)) {
         throw MLTConfigurationProblem(
-          ERS_HERE, get_name(), "Multiple subdetector types for ine SourceID not suported in trigger system!");
+          ERS_HERE, get_name(), "Multiple subdetector types for one SourceID not supported in trigger system!");
       }
       m_srcid_detid_map[sourceid] = detid;
     }
--- a/plugins/MLTModule.cpp
+++ b/plugins/MLTModule.cpp
@@ -34,12 +34,12 @@ MLTModule::MLTModule(const std::string& name)
   , m_run_number(0)
 {
   // clang-format off
-  //register_command("conf",   &MLTModule::do_configure);
+  register_command("conf",   &MLTModule::do_configure);
   register_command("start",  &MLTModule::do_start);
   register_command("stop",   &MLTModule::do_stop);
   register_command("disable_triggers",  &MLTModule::do_pause);
   register_command("enable_triggers", &MLTModule::do_resume);
-//  register_command("scrap",  &MLTModule::do_scrap);
+  register_command("scrap",  &MLTModule::do_scrap);
   // clang-format on
 }
 
@@ -50,7 +50,7 @@ MLTModule::decode_geoid(uint64_t _geoid_int)
 
   std::map<std::string, int> geoid;
 
-      // Extract stream_id (stored in the top 16 bits)
+  // Extract stream_id (stored in the top 16 bits)
   geoid["stream_id"] = (_geoid_int >> 48) & 0xFFFF;
 
   // Extract slot_id (stored in the next 16 bits)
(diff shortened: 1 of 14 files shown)


plugins/MLTModule.cpp
```diff
@@ -68,129 +67,158 @@ MLTModule::decode_geoid(uint64_t _geoid_int)
 void
 MLTModule::init(std::shared_ptr<appfwk::ConfigurationManager> mcfg)
 {
-  auto mtrg = mcfg->get_dal<appmodel::MLTModule>(get_name());
+  TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << get_name() << ": Entering init() method";
+
+  m_mtrg = mcfg->get_dal<appmodel::MLTModule>(get_name());
+  // Get the session to access the detector configuration
+  m_session = mcfg->session();
```
ArturSztuc, line 74: Why do we need this new global `m_session` object, rather than just calling `m_mtrg->session()` in the only place where it's called?

If we have `m_session`, then why not `m_inputs`, `m_outputs`, `m_configuration` etc?

plugins/MLTModule.cpp
```diff
@@ -68,129 +67,158 @@ MLTModule::decode_geoid(uint64_t _geoid_int)
...
       m_decision_output = get_iom_sender<dfmessages::TriggerDecision>(con->UID());
   }
 
-  // Get the session to access the detector configuration
-  auto session = mcfg->session();
-
-  hdf5libs::HDF5SourceIDHandler::source_id_geo_id_map_t geoidmap = hdf5libs::HDF5SourceIDHandler::make_source_id_geo_id_map(session);
-
+  hdf5libs::HDF5SourceIDHandler::source_id_geo_id_map_t geoidmap =
+    hdf5libs::HDF5SourceIDHandler::make_source_id_geo_id_map(m_session);
```
ArturSztuc, line 100: That's the only place `m_session` is being used. I guess the `auto session...` wasn't ideal either, but if change it, then why not to 

```suggestion
    hdf5libs::HDF5SourceIDHandler::make_source_id_geo_id_map(m_mtrg->session());
```
so we don't end up storing that object? Otherwise we could as well store far more (like the list of inputs etc.)

plugins/MLTModule.cpp
```diff
@@ -273,97 +312,97 @@ MLTModule::do_resume(const nlohmann::json& /*resumeobj*/)
...
-                    << ts << "/" << tt;
-    } else {
-      ers::warning(TriggerInhibited(ERS_HERE, m_run_number));
-      TLOG_DEBUG(1) << "The DFO is busy. Not sending a TriggerDecision with timestamp and type "
-                    << ts << "/" << tt;
-      m_td_inhibited_count++;
-      for ( const auto t : trigger_types ) {
-        ++get_trigger_counter(t).inhibited;
+      m_last_trigger_number++;
+      //        add_td(pending_td);
```
ArturSztuc: Funky formatting issue. If we're "improving" the formatting, then we should "improve"

plugins/MLTModule.hpp
```diff
@@ -136,8 +141,8 @@ class MLTModule : public dunedaq::appfwk::DAQModule
   // paused state, in which we don't send triggers
   std::atomic<bool> m_paused;
   std::atomic<bool> m_dfo_is_busy;
-  //std::atomic<bool> m_hsi_passthrough;
-  //std::atomic<bool> m_tc_merging;
+  // std::atomic<bool> m_hsi_passthrough;
+  // std::atomic<bool> m_tc_merging;
```
ArturSztuc, line 145: These are also odd. There's more of these, but I'll only highlight these...
Why was this even changed? This is not a "comment" but "commented-out code".

It would really be good if formatting changes were a separate PR from other code-changes PR. It's difficult to review all of this, the diff is all messed up.

src/TCProcessor.cpp
```diff
@@ -83,16 +89,19 @@ TCProcessor::stop(const nlohmann::json& args)
...
 }
 
 void
 TCProcessor::conf(const appmodel::DataHandlerModule* cfg)
 {
-  auto mtrg = cfg->cast<appmodel::TriggerDataHandlerModule>();	
-  if (mtrg == nullptr) {
+  TLOG_DEBUG(TLVL_ENTER_EXIT_METHODS) << "TCProcessor: Entering conf() method";
+
+  auto m_mtrg = cfg->cast<appmodel::TriggerDataHandlerModule>();
```
ArturSztuc: `m_` should only be reserved for class member variables!
Either remove `auto` (if `m_mtrg` is a member variable AND is being used in other functions) or remove `m_`.

src/trigger/TCProcessor.hpp
```diff
@@ -81,6 +84,8 @@ class TCProcessor : public datahandlinglibs::TaskRawDataProcessorModel<TCWrapper
   void add_requests_to_decision(dfmessages::TriggerDecision& decision,
                                 std::vector<dfmessages::ComponentRequest> requests);
 
+  const appmodel::TriggerDataHandlerModule* m_mtrg;
+
```
ArturSztuc: `m_mtrg` is only used in the function where it is being set, so why do we need it as a member variable?
Class member variables are there to be used throughout the class functions, not for _local_ variables.

ArturSztuc (commented):
Looks good to me, I only have minor comments (and main one being, changing formatting while making changes to the code makes PR long/difficult).

I'm still yet to test it, so this is not the end of my review :)

plugins/MLTModule.cpp
MRiganSUSX: Because the session is not a member of `m_mtrg` but instead of `mcfg` which is only available in the `init` stage. 
The other objects (inputs, outputs...) are all obtained from the `m_mtrg` which is the module configuration while the session is not part of this but a separate conf object.

plugins/MLTModule.cpp
MRiganSUSX: As explained before, session is not part of the configuration block for the MLTModule but a separete conf object. Therefore we need to 'stash' both the session and the module conf object.

plugins/MLTModule.cpp
MRiganSUSX: this is auto formatted by the clang script. I'll move it manually.

plugins/MLTModule.hpp
MRiganSUSX: This is also a result of the automated `dbt-clang-format.sh` script.
I agree that it makes it much harder to read what the actual code changes are but I don't know what the solution is. I'll give more details in the comment later.

MRiganSUSX:
Thanks for the initial review.
I have either addressed your comments in the code (and pushed one more commit) or explained the reasoning in the comments above. 

Regarding the formatting, it is more complicated. 
Ideally, everybody should be using the clang formatter script (`dbt-clang-format.sh`) and in that case, only the new code would be formatted with the script and therefore there would be no distinction (ie all the changes would be new code changes and nicely formatted). However, a lot of the trigger code is not (I think you were the last to do work on these modules ? :P). Therefore, it is only an issue if the code that the PR is modifying was not properly formatted before. Given we are expected to use the formatter (and produce properly formatted code) I have used it for my PR (and that caused the funky comments).
We could have the formatting applied as a separete PR each time, but that multiplies the number of PRs we make. 

I think the best thing to do now is to have a separate PR with the formatting only (applied across trigger code), and then apply the formatting for all future PRs which would only include the changes at that point. I'm happy to do the global format rework if you agree.

ArturSztuc (approved):
Tested both with integration tests and by hand with drunc, looking at logs at different stages, re-running conf after scrapping etc.

All pass👍🏻

PR #403, opened 2025-06-28
Small fix for replay app python script

MRiganSUSX:
There is a 'hack' in c++ tpreplay code that switches planes for APA1.
This hack also needs to be present in the python script preparing the configuration, so that the source ID numbers and plane numbers are matched up.


ArturSztuc (approved):
Tested and works, lgtm!

PR #409, opened 2025-09-02
Modified integtests to use the attribute_substitution data class instead of config_substitution. 

bieryAtFnal:
This is part of providing more configuration substitution options in integtests.

## Description

The changes in this PR make use of the new way of specifying configuration substitutions in integration tests, as described in DUNE-DAQ/daq-deliverables#181.

## Testing Suggestions

Please see the suggestions in the parent Issue, DUNE-DAQ/daq-deliverables#181.

### DIFF
--- a/integtest/change_rate_test.py
+++ b/integtest/change_rate_test.py
@@ -90,25 +90,25 @@ conf_dict.session = "changerate"
 conf_dict.tpg_enabled = False
 
 conf_dict.config_substitutions.append(
-    data_classes.config_substitution(
+    data_classes.attribute_substitution(
         obj_id=conf_dict.session,
         obj_class="Session",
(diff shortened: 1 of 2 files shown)


eflumerf:
I had a run failure for td_leakage_between_runs_test, is that one expected to be in the broken state?

bieryAtFnal:
Yes, there are pre-existing problems with a couple of the integtests in the trigger repo.  Here are the results from an integtest run **_without_** these changes, and we see the pre-existing problem with the `td_leakage_between_runs_test`.  So, I believe that the correct operation of the `change_rate_test` is a good indicator of the success of these changes.

Screenshot 2025-09-03 at 9 53 22 AM

eflumerf (approved):
Tests look good, with the noted exception of the previously-broken test.

PR #412, opened 2025-09-17
Patch/fddaq v5.4.x

jcfreeman2:
This will bring the busy-spin fix (https://github.com/DUNE-DAQ/trigger/pull/410) which necessitated `fddaq-v5.4.2` into `develop`. 

This can also be thought of as a fairly specific framing of "merge the patch branches for `fddaq-v4.4.2` into `develop`, since `fddaq-v4.4.2` only involved a change to this package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.12)
-project(trigger VERSION 2.5.0)
+project(trigger VERSION 2.5.1)
 
 find_package(daq-cmake REQUIRED)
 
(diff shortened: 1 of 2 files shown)


MRiganSUSX (approved):
I've run the `integtest_bundle` and all tests passed.

jcfreeman2:
Will proceed with merging. Please note the reason for the CI action failure is that since the target branch is develop, this action is trying to build `patch/fddaq-v5.4.x` branches against the nightly, even though the interfaces have diverged; this doesn't indicate a problem with the PR itself but rather suggests there may be ways to make the CI action itself more robust against false outcomes.

PR #416, opened 2025-10-07
Delayed postprocessing timeout logic

denizergonul:
Trigger has to define a policy on what to do in case of a timeout. For this, we introduced a new function in `DataHandlingModel` that will be overriden by `TriggerDataHandlingModel`. The function is called by postprocess scheduler.

https://github.com/DUNE-DAQ/datahandlinglibs/pull/104


denizergonul:
@alessandrothea could you help me fill in the DUNE DAQ tracking properties?

PR #417, opened 2025-10-09
PreconfiguredTriggerModule for SNB Transform demonstrator

eflumerf:
This PR adds a PreconfiguredTriggerModule TC generator. It generates TCs based on PreconfiguredTriggerModuleTrigger objects linked to the configuration. This is used for the SNB Transform Demonstrator, https://github.com/DUNE-DAQ/daq-deliverables/issues/186, to generate the triggers used to read out the binary data through the DUNE-DAQ system.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,60 @@
+name: Bug Report
+description: Report a bug in the system
+title: "[Bug]: "
+labels: ["bug"]
+projects: ["DUNE-DAQ/5"]
+body:
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: Give a clear and concise description of the bug, included what you expected to happen and what the actual behavior was.
+      placeholder: "What happened?"
+    validations:
+      required: true
+  - type: textarea
+    id: reproduce
+    attributes:
+      label: Steps to reproduce
+      description: Give a clear set of steps to reproduce the error or how you saw the error.
+      placeholder: |
+        1. Go to...
+        1. Setup...
+        1. Run...
+        1. See error...
+    validations:
+      required: true
+  - type: textarea
+    id: version
+    attributes:
+      label: Version and environment
+      description: Give the dunedaq version and environment in which you saw the error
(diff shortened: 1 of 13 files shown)


eflumerf:
From the SNB Demonstrator presentation (linked from the main Issue):

* For simple SNB trigger tests with existing files, the easiest way to trigger is to read the first and last timestamps from the frames in the file, and create a synthetic trigger covering that time range.
* PreconfiguredTriggerModule is a TC generator like RandomTCMakerModule
* It can be configured with one or more triggers to send, and the time delay the module should insert between triggers (i.e. for processing the trigger)
* Uses TCReadoutMap to configure TC type and window extents (for SNB transform test, kSupernova and 0 offsets are used)

alessandrothea:
Thanks for the PR. 
The code looks largely ok at this stage. My main comments are about the naming:
- Up so far modules that create modules are called "makers". While we may change in the future I would stick to it for now
- `PreconfiguredTrigger` is quite confusing: if I understand correctly, the defining characteristic of this module is that it injects trigger candidates at pre-determined times, right?

eflumerf:
Would you prefer `PreconfiguredTCMakerModule`?

The defining characteristic of this module is that it makes TCs with pre-defined start and end times, from the configuration. This is intended for retrieving previously-recorded data from a file.

alessandrothea:
Thanks, it goes in the right direction.
I still find the use of `Preconfigured` confusing, modules are configured, TCs are not.
Maybe `PredefinedTC` or `FixedTimeTC` or... would make the purpose of the module clearer.

Just a clarification: once the list of configured triggers is exausted, no more triggers are sent and the module effectively stops, right?

One more thing: I would not use of the `TCReadoutMap` object to specify the window extent and rather add the variables to the conf class directly. The readout map has little to do with candidate creation -  I am removing it from the RCTM as well at the next chance.

eflumerf:
`FixedTimeTCMaker` seems to communicate the intent pretty well. Yes, once the configured triggers are done, the module does nothing else. I'll remove the window adjustments entirely, as they were really a hack to try to get things working with the DefaultRequestHandler, but now the SNB transform has its own SNBRequestHandler.

andrewmogan (approved):
LGTM.

See comments in [daq-deliverables #189](https://github.com/DUNE-DAQ/daq-deliverables/issues/186) for tests performed.

PR #418, opened 2025-10-17
Updated the WIBEth binary file used in the td_leakage_between_runs_test

bieryAtFnal:
## Description

DUNE-DAQ/daqsystemtest#245 requests that we update our regression tests, etc. to use recent data-replay files from the _assets_ system.

In this repo, that translated to changing the checksum of the asset file that is used in the `td_leakage_between_runs_test`.  The new file is `wib_link_67.bin`, which was recently created by Eric.

*** Please Note that we modified the desired change in this repo.  Instead of changing the default value of the frames file, we simply removed the setting of the frames file value.

## Testing Suggestions

There are suggestions in DUNE-DAQ/daqsystemtest#245 for testing all of the asset-file-update changes together.  

However, the `td_leakage_between_runs_test` is currently broken, so the review of this PR should simply visually check that the new checksum is valid.  The test will be fixed soon, but not necessarily part of this PR.

### DIFF
--- a/integtest/td_leakage_between_runs_test.py
+++ b/integtest/td_leakage_between_runs_test.py
@@ -74,7 +74,7 @@ conf_dict["hsi"]["use_timing_hsi"] = False
 
 conf_dict["readout"]["data_files"] = []
 datafile_conf = {}
-datafile_conf["data_file"] = "asset://?checksum=e96fd6efd3f98a9a3bfaba32975b476e" # WIBEth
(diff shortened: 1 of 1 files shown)


eflumerf (approved):

eflumerf:
Question: If this is the default frame file, should it be specified at all?

bieryAtFnal:
As suggested by Eric, I removed the specification of the data_file in favor of a comment that describes a bit about what happens in the background with defaults.

PR #419, opened 2025-10-21
New RTCM configuration

alessandrothea:
Address the RTCM limitations in #415.
The candidate time is shifted by `candidate_offset_ts` (signed integer, leaving the option to inject candidates into the future if needed), and the window is define by `candidate_window_before_ts`/`candidate_window_after_ts`

The use of `TCReadoutMap` for the window definition is discontinued (since it belongs to a different domain)

Requires associated changes in `appmodel`
https://github.com/DUNE-DAQ/appmodel/pull/259

PR #422, opened 2025-11-25
Revert "New RTCM configuration"

wesketchum:
Reverts DUNE-DAQ/trigger#419

Reverting due to failures of integtests in daqsystemtest (see above), and also suspect we need to look at other places where the config needs changes/updates (dfmodules integtests, dundedaq_conf_reader, ...)


bieryAtFnal (approved):
I have run regression tests from `daqsystemtest, dfmodules, trigger, asiolibs`, and `crtmodules` with these changes and the corresponding changes in `appmodel` and `daqsystemtest`, and the tests that were expected to pass did so.  (My software area started with the v5.5.0 RC2 build, included the repos that have changes for RC3, and included these changes.)

Issue #428, opened 2026-03-11
[Feature]: New and replacement integration / regression tests


bieryAtFnal:
### Description

Two of the three existing integtests in the `trigger` repo are broken. This has been true for a while, and I am going to remove them, for now.

However, the functionality that they once tested would still be good to have.  So, I'm filing this Issue to remind us to come back to these tests someday.

- the `tc_time_outside_window_test.py` test attempts to validate the expected behavior of TriggerCandidate fragments when the TriggerDecision request window doesn't include the TC time.  Unfortunately, at the moment, it is not clear to me how we expect TC fragments to be populated in various conditions, and I'm not sure of the current way (or ways) that we specify TC times versus other times.
- the `td_leakage_between_runs_test.py` test attempted to validate the graceful stopping of TriggerDecision messages between the MLT and the DFO.  There are comments in this [this PR](https://github.com/DUNE-DAQ/trigger/pull/249). The stumbling block to getting this integtest working again now is that it is not clear how we would configure the conditions that are likely to demonstrate the problem

In summary, it would be good to bring back these tests someday.

Independent of that, we can imagine other tests that would be useful.  Such as ones that validate the behavior of TC merging.

### Potential impact radius

Small/Isolated

### Reason for change

Validating expected functionality is a good thing

MRiganSUSX:
For future reference, there are some trigger related integtests in daqsystemtest, for example: [trigger_bitwords_test.py](https://github.com/DUNE-DAQ/daqsystemtest/blob/develop/integtest/trigger_bitwords_test.py) and [tpreplay_test.py](https://github.com/DUNE-DAQ/daqsystemtest/blob/develop/integtest/tpreplay_test.py). 
Especially `trigger_bitwords_test.py` covers several interesting cases, one of them being TC merging.

PR #429, opened 2026-03-11
removed the broken integtests, for now...

bieryAtFnal:
These are `tc_time_outside_window_test.py` and `td_leakage_between_runs_test.py,`.

# Description

Please see DUNE-DAQ/trigger#428 for more information about the broken tests that are being removed.  

I am removing them so that we can stop cluttering up regression test results with ones that are known to be broken.


MRiganSUSX (approved):
Thanks @bieryAtFnal.
lgtm

